PR #41, opened 2018-09-12
doc comment fix; resolves #7

### LINKED ISSUE #7: Doc comment on `data TDef` is rejected by the parser
https://github.com/typedefs/typedefs/blob/master/src/Typedefs.idr#L13

It says `-- |||` now but it ought to be `|||`; not sure why this is happening, or if this still happens for Idris 1.3.

tg-x:


epost:
Ah, and whitespace fixes; yay emacs! 👍

PR #43, opened 2018-09-12
build: separate pkg dirs, remove unnecessary deps

tg-x:
- remove contrib dependency, since it does not seem to be necessary
- move packages to separate source directories
   - this means building the executables now depend on an installed typedefs lib, since it's not possible to specify multiple source or lib dirs
   - the examples can still be run via the REPL without installing the lib (see README)


tg-x:
cc @alex

PR #45, opened 2018-09-17
nix: enable tests

tg-x:
enable checks when building with nix, such that travis would also run the test suite


epost:
Thanks @tg-x! 👍 👍

Btw, it's a good idea to mention me, or to assign me, or to request a review with that thingie in the top right corner, or do some combination of the above.

PR #55, opened 2018-10-02
Add test case for parsing TName terms. #54

epost:


clayrat:
Ok I guess I'll just do it myself with a next PR for tests.

PR #64, opened 2018-10-02
generate haskell types instead of showing in the JS app

clayrat:
(I don't think we have an issue for this)


clayrat:
@epost Should we merge this? :)

clayrat:
@epost done

epost:
@clayrat 
2kcge9

epost:
@clayrat lovely, thanks! 👍

PR #74, opened 2018-11-15
ReasonML backend

kjekac:
Passes "the same" tests as the Haskell backend, plus an additional one that both of them failed (`parametricSynonym`). In module `Types` I've also added the functions `uppercase, lowercase : Name -> Name`, meant to aid in the compliance with each language's concrete syntax.


src/Backend/ReasonML.idr
```diff
@@ -0,0 +1,95 @@
...
+import Control.Monad.State
+
+import Types
+import Typedefs
+
+%default partial
+%access public export
+
+Env : Nat -> Type
+Env k = Vect k (Either String String) -- left=free / right=bound
```
clayrat: we should probably extract this repeating stuff into a common module

src/Backend/ReasonML.idr
```diff
@@ -0,0 +1,82 @@
...
+                                in if freeVars == "" then lowercase nam else lowercase nam ++ parens freeVars
+makeType     _ (TName nam _) = lowercase nam
+
+makeDefs : Env n -> TDef n -> State (List Name) String
+makeDefs _ T0           = pure ""
+makeDefs _ T1           = pure ""
+makeDefs e (TProd xs)   = map concat $ traverse (makeDefs e) xs
+makeDefs e (TSum xs)    = map concat $ traverse (makeDefs e) xs
+makeDefs _ (TVar v)     = pure ""
+makeDefs e (TMu nam cs) =
```
epost: Is the reason for abbreviating `name` to `nam` that t would cause shadowing?

src/Backend/ReasonML.idr
```diff
@@ -0,0 +1,82 @@
...
+  mkArg : Env (S n) -> (Name, TDef (S n)) -> String
+  mkArg _ (cname, T1)       = cname
+  mkArg e (cname, TProd xs) = uppercase cname ++ parens (withSep ", " (makeType e) xs)
+  mkArg e (cname, ctype)    = uppercase cname ++ parens (makeType e ctype)
+makeDefs e (TName nam body) = 
+  do st <- get 
+     if List.elem nam st then pure "" 
+       else let 
+          freeVars = getFreeVars e
+          typeName = if freeVars == "" then lowercase nam else lowercase nam ++ parens freeVars
```
epost: Same `... ++ if ...` thing as above.

src/Backend/ReasonML.idr
```diff
@@ -0,0 +1,82 @@
...
+           put (nam :: st)
+           pure $ res ++ "\ntype " ++ typeName ++ " = " ++ makeType e body ++ ";\n"
+  
+
+-- generate type body, only useful for anonymous tdefs (i.e. without wrapping Mu/Name)
+generateType : TDef n -> String
+generateType {n} = makeType (freshEnv n)
+
+-- generate data definitions
+generate : TDef n -> String
```
epost: More of a general question: would it be possible to start formulating some sort of interface for functions such as this one?

examples/Examples.idr
```diff
@@ -5,6 +5,7 @@ import Types
 import Typedefs
 import TermCodec
 import Backend.Haskell
+import Backend.ReasonML
 
 ------ example: bits -
```
epost: Why does this comment start with 5 dashes followed by a lowercase letter, as opposed to the others, which look like `-- Example:`?

src/Backend/ReasonML.idr
clayrat: I think this is just mirroring what we did in the haskell one, and the reason there was basically following the academic style!

examples/Examples.idr
kjekac: I was wondering the same thing when I was editing the file, but figured that you had your reasons. Can absolutely edit if desired!

src/Backend/ReasonML.idr
kjekac: Yes I think we should, will start trying things out!

src/Backend/ReasonML.idr
kjekac: I followed the conventions from the Haskell backend, renaming to `name` doesn't cause errors so I'll change it.

src/Backend/ReasonML.idr
```diff
@@ -0,0 +1,82 @@
...
+makeDefs e (TProd xs)   = map concat $ traverse (makeDefs e) xs
+makeDefs e (TSum xs)    = map concat $ traverse (makeDefs e) xs
+makeDefs _ (TVar v)     = pure ""
+makeDefs e (TMu nam cs) = 
+  do st <- get 
+     if List.elem nam st then pure "" 
+      else let
+         freeVars = getFreeVars e
+         dataName = if freeVars == "" then nam else nam ++ parens freeVars
+         newEnv = Right (if freeVars == "" then dataName else dataName) :: e
```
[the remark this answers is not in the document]
kjekac: Correct, stupid, fixing!

src/Backend/ReasonML.idr
kjekac: No because `parens` gives us at least `()` back, but you're right, that's hardly desirable behavior; fixing.

src/Backend/ReasonML.idr
kjekac: The problem is that because we're generating a simple `String`, there's really nothing we can capture about the target language on the type level, unless we start generating syntax trees or something similar. (Which seems like it could be a good idea, but a slightly bigger decision than I'm comfortable making still...)

epost: Actually, it seems like a good start to say which `-> String` functions we have in the first place. That would also at least give us a good place to add some doc comments. Maybe some of them should be `-> Maybe String` instead? I'm not sure. Anyway, if we want to get more detailed at some point, we can always do that.

epost:
Thanks Jack, looks great! :D Apart from the merge conflicts this is good to go afaic. If you're going to fix those conflicts, perhaps have a look at the bottom of two of the files (2 vs. no newlines) and fix those as well for extra unicorn sparkles.

PR #76, opened 2018-11-16
Haskell backend handles parametric type synonyms correctly

kjekac:


src/Backend/Utils.idr
```diff
@@ -17,4 +17,8 @@ freshEnv : (n: Nat) -> Env n
 freshEnv n = unindex {n} (\f => Left ("x" ++ show (finToInteger f)))
 
 withSep : String -> (a -> String) -> Vect k a -> String
-withSep sep fn xs = concat $ intersperse sep $ map fn xs
\ No newline at end of file
+withSep sep fn xs = concat $ intersperse sep $ map fn xs
+
+getFreeVars : (e : Env n) -> Vect (fst (Vect.filter Either.isLeft e)) String
+getFreeVars e with (filter isLeft e) 
+  getFreeVars e | (p ** v) = map (either id (const "")) v
```
clayrat: You can even remove the `getFreeVars e` part in this line, since none of the original params get changed by `with`

PR #83, opened 2018-11-29
Let-style parsing

### LINKED ISSUE #61: Let-bindings
We should probably introduce some aliasing/let binding syntax, so that we can define them in parts and referred back to, like in the `ListNat` test.

Syntax suggestions? Shall we introduce list literals (which i don't particularly like, but wth):
```
(let [ (name Maybe (+ 1 (var 0))
     , (mu ...)
     , ...
     ]
     (....tdef...)
)
```
Does this make any sense? :P I'm half-tempted to break sexp syntax where and go for a list of definitions like

```
foo = (+ 1 1)
bar = (...)
```

Actually, how should this interact with naming through `TName` and `TMu`?

clayrat:
Closes #61

### DIFF
--- a/src/Parse.idr
+++ b/src/Parse.idr
@@ -92,7 +92,7 @@ tdef =
         , map (\n => (S n ** TVar $ last {n})) $
             parens (rand (withSpaces (string "var")) (withSpaces decimalNat))
         , randbindm
-            (guardM {b=(String, (n ** TDef n))} 
+            (guardM {b=(String, (n ** TDef n))}
                     (\(nam, nel) =>
                      let vs : Vect (length nel) (n : Nat ** (String, TDef n)) =
                            -- push names under sigma to fit in VMax
--- a/src/Parse.idr
+++ b/src/Parse.idr
@@ -131,14 +131,24 @@ tdef =
                 (map {a=Parser' _} withSpaces rec)
                 (map {a=Parser' _} (nelist . withSpaces) rec)))
 
+||| Parse a sequence of TDefs and return the last one that parsed, accumulating
+||| and substituting named entries in the process
 tdefRec : All (Parser' (n ** TDef n))
 tdefRec = fix _ $ \rec => map (\(a, ma) => fromMaybe a ma) $ andopt tdef rec
+
+||| Parse a sequence of TDefs and return a non-empty list of all results,
+||| accumulating and substituting named entries in the process
+tdefNEL : All (Parser' (NEList (n ** TDef n)))
+tdefNEL = nelist tdef
 
 parseTDef : String -> Maybe (n : Nat ** TDef n)
 parseTDef str = parseMaybe str tdefRec
 
+parseTDefs : String -> Maybe (NEList (n : Nat ** TDef n))
+parseTDefs str = parseMaybe str tdefNEL
+
 parseThenShowTDef : String -> String
-parseThenShowTDef str = show $ parseTDef str
+parseThenShowTDef = show . parseTDef
 
 parseThenStrFun : String -> ((n ** TDef n) -> String) -> String
 parseThenStrFun str fn = maybe ("Failed to parse '" ++ str ++ "'.") fn $ parseTDef str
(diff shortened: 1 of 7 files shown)


src/Parse.idr
```diff
@@ -38,56 +41,101 @@ fromVMax {m} vm = go lteRefl vm
...
+
+andbindm : Monad mn =>
+  All (Parser mn p a :-> Cst (a -> mn b) :-> Parser mn p (a, b))
+andbindm p f = MkParser $ \mlen, ts => do ra <- runParser p mlen ts
+                                          b <- f (Value ra)
+                                          pure $ map (flip MkPair b) ra
+
+landbindm : Monad mn =>
+  All (Parser mn p a :-> Cst (a -> mn b) :-> Parser mn p a)
+landbindm p f = map fst (andbindm p f)
```
fredrikNordvallForsberg: Is there ever an advantage to use `landbindm p f` instead of just `p`? In case you want to allow `f` the possibility to fail, or cause a side-effect?

src/Parse.idr
```diff
@@ -38,56 +41,101 @@ fromVMax {m} vm = go lteRefl vm
...
+                          (fromVMax vx))
+       ) $
+       parens (rand (withSpaces (char sym))
+          (map2 {a=Parser' _} {b=Parser' _}
+                (\p, q => and p q)
+                (map {a=Parser' _} withSpaces rec)
+                (map {a=Parser' _} (nelist . withSpaces) rec)))
+
+tdefRec : All (Parser' (n ** TDef n))
+tdefRec = fix _ $ \rec => map (\(a, ma) => fromMaybe a ma) $ andopt tdef rec
```
fredrikNordvallForsberg: If I read this correctly, this will parse a sequence of `TDef`s, and return the last one (that parsed)? Perhaps there would be some value in also defining the function that returns a list of `Tdef`s?

src/Test/Parse.idr
```diff
@@ -48,6 +48,10 @@ testSuite = spec $ do
 
     "(+ 1 1 0 (* 1 0))" `itShouldParseAs` "Just (0 ** (1 + 1 + 0 + (1 * 0)))"
 
+    "(mu nat (Z 1) (S (var 0))) (name mnat (+ 1 nat)) (mu listmnat (nil 1) (cons (* mnat (var 0))))"
+      `itShouldParseAs`
+        "Just (0 ** listmnat = mu [nil: 1, cons: ((1 + nat = mu [Z: 1, S: {0}]) * {0})])"
```
fredrikNordvallForsberg: Just a prettyprinting nitpick, but I think it would aid readability to put `nat = mu [Z: 1, S: {0}]` in brackets. Perhaps `:=` would be more readable than `=` too, e.g.
```
"Just (0 ** listmnat := mu [nil: 1, cons: ((1 + (nat := mu [Z: 1, S: {0}])) * {0})])"
```
? I hope that's an easy change, otherwise it's probably not worth the effort.

src/Parse.idr
clayrat: Yeah, I added it just for the sake of completeness, with side effects in mind. I'm going to move this section to `TParsec` proper after this gets merged, so maybe Guillaume will have to say something about it :)

clayrat:
@fredrikNordvallForsberg yeah, I think the root cause for slowness is that fact that every combinator used adds a bunch of new independent constraints to the scope, even though they all are satisfied by the same handful of implementations that you have; and the search probably is smth like quadratic in the number of constraints :) This will be worth looking at in Blodwen.

src/Parse.idr
epost: Would it make sense to add @fredrikNordvallForsberg's remark as a doc comment?

src/Test/Parse.idr
epost: Not sure if you are going to do this @clayrat but let me know if I need to merge anything.

clayrat: Yes, gonna push this tomorrow

clayrat:
@epost all done I think

Issue #81, opened 2018-11-24
Specialisation to language specific 'primitives'


wires:
We can for instance define `bit : 1 + 1` and then `nibble : bit * bit * bit * bit` and then `byte : nibble * nibble`.

If when we translate this into a target language, we would like the interpretation of `byte` to be `Char`, not a big product of `Either Unit Unit`.

Thus, when we generate de/encoding functions for a target language, this should consider the type specialisations for this language. A naive approach could be that we keep a table of specialisations, that map hashes of serialised typedefs into the specialised type of the target language.

| Hash                                                                    | Target  | Special Type |
|-------------------------------------------------------------------------|---------|--------------|
| sha256-2413fb3709b05939f04c| haskell | Char         |
| sha256-e92f7d0897fc2596f9ad0 | ocaml   | Bytes        |
| etc                                                                        |         |              |

For example:

```
(type byte) == Char
decode_byte : Buffer -> Maybe Char
encode_byte : Char -> Buffer

(type bit) == (type (1 + 1))
decode_bit : Buffer -> Maybe (Either Unit Unit)
encode_bit: Either Unit Unit -> Buffer
```

If a (type, target language) pair doesn't have an entries, the normal, unspecialised type should be used.

It seems we should only generate codecs for closed typedefs: such as `List Byte`, which interacts with [application #79](https://github.com/typedefs/typedefs/issues/79) (as we need to apply `List a` to `Byte`.

wires:
cc @epost @clayrat @fredrikNordvallForsberg @kjekac wrote up some basic remarks about the "specialisation"

PR #85, opened 2018-12-06
Specialise machinery

fredrikNordvallForsberg:
This addresses #81, in a backend-independent manner, i.e. backend writers need not concern themselves with specialisation. (This means a performance cost at compile time, since we do at least two passes over the `TDef`, but we probably won't compile massive typedefs anyway, and a clever compiler could fuse the two passes, perhaps. (Of course, I highly doubt Idris has a clever compiler.))

So far the work is only done for types, not term encoding/decoding, since that is missing in general in the backends so far.

For discussion:
* Currently the order in which specialisations are listed can influence the output, for instance consider
```
bit = Tsum [T1, T1]
byte = TProd [bit, bit, bit, bit, bit, bit, bit, bit]
twoBytes = TName "TwoBytes" (TProd [byte, byte])
```
and specialisations `sBit : bit |-> Bool` and `sByte : byte |-> Char`. Generating Haskell code for twoBytes with different specialisations results in the following:

Specialisation | Outcome
-------------------|-------------
| []    | `type TwoBytes = ((Either () (), Either () (), Either () (), Either () (), Either () (), Either () (), Either () (), Either () ()), (Either () (), Either () (), Either () (), Either () (), Either () (), Either () (), Either () (), Either () ()))` |
| [sBit]    | `type TwoBytes = ((Bool, Bool, Bool, Bool, Bool, Bool, Bool, Bool), (Bool, Bool, Bool, Bool, Bool, Bool, Bool, Bool))` |
| [sByte]    | `type TwoBytes = (Char, Char)` |
| [sByte, sBit]    | `type TwoBytes = (Char, Char)` |
| [sBit, sByte]    |  `type TwoBytes = ((Bool, Bool, Bool, Bool, Bool, Bool, Bool, Bool), (Bool, Bool, Bool, Bool, Bool, Bool, Bool, Bool))` |

Why? If we do `[sBit, sByte]`, we first translate `twoBytes` into
```
TName "TwoBytes" (TProd [TProd [Bool, Bool, Bool, Bool, Bool, Bool, Bool, Bool],
                         TProd [Bool, Bool, Bool, Bool, Bool, Bool, Bool, Bool]])
```
with `sBit`, which means that `sByte` no longer matches. Is this a problem? It seems hard to address in general. Perhaps it is the user's fault for giving conflicting specialisations?


clayrat:
```
Parser.idr:7:4:
  |
7 |   [_, str] <- getArgs
  |    ^
When checking right hand side of Main.case block in main at Parser.idr:7:15-21 with expected type
        IO ()
When checking an application of function Prelude.Interactive.putStrLn:
        Can't disambiguate since no name has a suitable type: 
                Prelude.List.++, Prelude.Strings.++
```

PR #86, opened 2018-12-10
prettyprint and interface for backends

kjekac:
The central idea here is to split the backend into two parts and create an interface for these two parts, as can be seen here:
```idris
-- Interface for codegens. lang is a type representing (the syntactic structure of)
-- a type declaration in the target language.
interface Backend lang where
	-- Given a TDef and a matching environment, generate a list of type definitions
	-- in the target language.
  generateDefs : Env n -> TDef n -> List lang

  -- Given a type definition in the target language, generate its code.
  generateCode : lang -> Doc
```

Given this interface, we can expose a function `generate : Backend lang => {n: Nat} -> TDef n -> Doc` which can be called as e.g. `generate {lang=Haskell} td`.

# Issues

## Orthogonal approach to PR #85 
PR #85 uses a `record` instead of an `interface` to generalize backends. There are also some diffences in the functions that are expected to be implemented. I can see merits in both.

For example, the interface used in this PR only cares about actual type definitions and completely disregards how a single type signature is generated, which to me seems like a desirable abstraction. Or do we actually want to generate a single type signature anywhere outside of the backend itself? I imagine that the generation of (de)serializers will be part of the backends, too.

On the other hand, the record used in PR #85 disregards the syntactic structure of the target language, which also seems like a desirable abstraction – I have been a bit uncomfortable with the intermediate step my solution introduced.

In this PR, I also assumed that `freshEnv` can be reused across backends but realize that due to syntactic differences this might not be the case.

Given this, I suggest that we unify the two approaches into something like the following:

```idris
record Backend where
  constructor MkBackend
  ||| Generate data definitions
  generateDefs : Env n -> TDef n -> Doc
  freshEnv : (n : Nat) -> Env n
```

## Code repetition
I am not completely happy with the amount of code repetition between the Haskell and ReasonML backend; I suspect that a lot can be extracted into language agnostic code. However, given the high similarities of the two languages' syntaxes, I want to avoid doing this at this moment. If this is a priority, I believe we should add a backend for a third, preferably fairly different, language, in order to see clearly which parts that are in fact irrelevant to the target language, and which parts just happen to be similar between Haskell and ReasonML. This should also wait until after we have solved the issue above.


kjekac:
Realize that this is WIP but I would like feedback on the first issue I mention above before I go ahead and pick a solution approach.

src/Backend/Haskell.idr
```diff
@@ -1,68 +1,107 @@
...
+  HVar   :         Name               -> HType
+  HParam : Name -> Vect k HType       -> HType
+
+data Haskell : Type where
+  Synonym : Decl -> HType                -> Haskell
+  ADT     : Decl -> Vect k (Name, HType) -> Haskell
 
-nameWithParams : Name -> Env n -> String
-nameWithParams name e = withSep " " id (uppercase name::map lowercase (getFreeVars e))
+-- Render a name applied to a list of arguments exactly as written (arguments need to be previously parenthesized, if applicable).
```
epost: I think many of the comments could (and probably should) be turned into doc comments.

src/Backend/Haskell.idr
```diff
@@ -1,68 +1,107 @@
...
 
-%default partial
+import Data.Vect
+
+%default total
 %access public export
 
-guardPar : String -> String
-guardPar str = if any isSpace $ unpack str then parens str else str
+data HType : Type where
```
epost: I think `HType` could do with a doc comment. And perhaps a different name? From seeing this now for the 1st time I'm assuming this means 'Haskell type' (and not e.g. heterogeneous type) so perhaps `HaskellType` is too long (?), and single letters are admittedly noice, but how do we feel about eg `HsType`? Just throwing this out there...

src/Backend/Haskell.idr
```diff
@@ -1,68 +1,107 @@
...
-makeType : Env n -> TDef n -> String
-makeType     _ T0             = "Void"
-makeType     _ T1             = "()"
-makeType {n} e (TSum xs)      = tsum xs
+mutual
+  -- Render a type signature as Haskell source code. 
+  renderType : HType -> Doc
+  renderType H0                   = text "Void"
+  renderType H1                   = text "()"
+  renderType p@(HProd xs)         = tupled . toList $ map (assert_total renderType) xs
```
epost: `p` is not used

src/Backend/Utils.idr
```diff
@@ -1,24 +1,64 @@
...
+  | (p ** v) = map (either id (const "")) v
+
+-- 
+vsep2 : List Doc -> Doc
+vsep2 = vsep  . punctuate line
+
+print : Doc -> String
+print = toString 1 80
+
+-- Interface for codegens. lang is a type representing (the syntactic structure of)
```
epost: I think `lang` could do with verbatim markup (or whatever other markup is useful) in the doc comment version of this.

src/Backend/Utils.idr
```diff
@@ -1,24 +1,64 @@
...
+-- a type declaration in the target language.
+interface Backend lang where
+	-- Given a TDef and a matching environment, generate a list of type definitions
+	-- in the target language.
```
epost: `TDef` markup?

src/Backend/Utils.idr
```diff
@@ -1,24 +1,64 @@
...
+  generateDefs : Env n -> TDef n -> List lang
+
+  -- Given a type definition in the target language, generate its code.
+  generateCode : lang -> Doc
+
+-- Generate code in a specific language for a type definition and all its dependencies.
+-- Needs to be called with {lang}
```
epost: `{lang}` markup?

epost:
@kjekac As a general comment, I think it would be good to think about how we could name these things in such a way that it's clear whether we're dealing with types or terms. E.g. 'generateDefs` seems a trifle ambiguous (disregarding its type, that is).

kjekac:
This should be mergable now as per our discussion in the meeting.

PR #87, opened 2018-12-17
Specialise machinery uses the Backend interface

kjekac:
This is a merge of #86 into #85. 

Given that we are generally satisfied with those two PRs, I think this one should be very uncontroversial.


src/Backend.idr
```diff
@@ -0,0 +1,102 @@
...
+getUsedIndices (TSum xs)  = assert_total $ concatMap getUsedIndices xs
+getUsedIndices (TProd xs) = assert_total $ concatMap getUsedIndices xs
+getUsedIndices (TVar i)   = [i]
+getUsedIndices (TMu _ xs) = assert_total $ concatMap ((concatMap weedOutZero) . getUsedIndices . snd) xs
+  where weedOutZero : Fin (S n) -> List (Fin n)
+        weedOutZero FZ     = []
+        weedOutZero (FS i) = [i]
+getUsedIndices (TName _ t) = getUsedIndices t
+
+||| Filter out the entries in an Env that is referred to by a TDef.
```
epost: Markup for `Env` and `TDef`?

src/Backend.idr
```diff
@@ -0,0 +1,102 @@
...
+getUsedVars : Env n -> (td: TDef n) -> Env (length (getUsedIndices td))
+getUsedVars e td = map (flip index e) (fromList $ getUsedIndices td)
+
+||| Interface for codegens. lang is a type representing (the syntactic structure of)
```
epost: `lang`

src/Backend.idr
```diff
@@ -0,0 +1,102 @@
...
+||| Vertically concatenate a list of documents with two newlines (i.e. one empty line) as separator.
+vsep2 : List Doc -> Doc
+vsep2 = vsep . punctuate line
+
+||| Get the free names from the environment.
+getFreeVars : (e : Env n) -> Vect (fst (Vect.filter Either.isLeft e)) String
+getFreeVars e with (filter isLeft e)
+  | (p ** v) = map (either id (const "")) v
+
+||| Get a list of the de Brujin indices that are actually used in a TDef.
```
epost: `TDef`

src/Backend/Haskell.idr
```diff
@@ -1,68 +1,127 @@
...
-        in
-       do res <- map concat $ traverse {b=String} (\(_, bdy) => makeDefs newEnv bdy) cs 
-          put (name :: st)
-          pure $ res ++ "\ndata " ++ dataName ++ " = " ++ args ++ "\n"
+  renderConstructor : (Name, HsType) -> Doc
+  renderConstructor (name, HsUnit)     = renderApp name []
+  renderConstructor (name, HsTuple ts) = renderApp name (map guardParen ts)
+  renderConstructor (name, params)     = renderApp name [guardParen params]
+
+||| Generate a Haskell type from a TDef.
```
epost: `TDef`

src/Backend/Haskell.idr
```diff
@@ -1,68 +1,127 @@
...
-  mkArg _ (cname, T1)       = cname
-  mkArg e (cname, TProd xs) = cname ++ " " ++ withSep " " (makeType e) xs
-  mkArg e (cname, ctype)    = cname ++ " " ++ makeType e ctype
-makeDefs e (TName name body) = 
+  hsParam : Decl -> HsType
+  hsParam (MkDecl n ps) = HsParam n (map HsVar ps)
+makeType e td@(TMu name _)   = HsParam name . map HsVar $ getFreeVars (getUsedVars e td)
+makeType e td@(TName name _) = HsParam name . map HsVar $ getFreeVars (getUsedVars e td)
+
+||| Generate Haskell type definitions from a TDef, including all of its dependencies.
```
epost: `TDef`

src/Backend/ReasonML.idr
```diff
@@ -1,73 +1,135 @@
...
+                                 |++| equals |++| hsep (punctuate (text " |") (toList $ map renderConstructor cases))
+                                 |+| semi
+  where
+  renderConstructor : (Name, RMLType) -> Doc
+  renderConstructor (name, RMLUnit)     = renderApp (uppercase name) []
+  renderConstructor (name, RMLTuple ts) = renderApp (uppercase name) (map renderType ts)
+  renderConstructor (name, t)           = renderApp (uppercase name) [renderType t]
+
+
+||| Generate a ReasonML type from a TDef.
```
epost: `TDef`

src/Backend/ReasonML.idr
```diff
@@ -1,73 +1,135 @@
...
-makeDefs e (TProd xs)    = map concat $ traverse (makeDefs e) xs
-makeDefs e (TSum xs)     = map concat $ traverse (makeDefs e) xs
-makeDefs _ (TVar v)      = pure ""
-makeDefs e (TMu name cs) = 
+  rmlParam : Decl -> RMLType
+  rmlParam (MkDecl n ps) = RMLParam n (map RMLVar ps)
+makeType     e td@(TMu   name _) = RMLParam name . map RMLVar $ getFreeVars (getUsedVars e td)
+makeType     e td@(TName name _) = RMLParam name . map RMLVar $ getFreeVars (getUsedVars e td)
+
+||| Generate ReasonML type definitions from a TDef, includig all of its dependencies.
```
epost: `TDef`

src/Test.idr
```diff
@@ -0,0 +1,57 @@
...
+aplusbpluscplusd : TDef 4
+aplusbpluscplusd = TName "aplusbpluscplusd" $ TSum [TVar 0, TVar 1, TVar 2, TVar 3]
+
+oneoneoneone : TDef 0
+oneoneoneone = TName "oneoneoneone" $ TSum [T1, T1, T1, T1]
+
+listByte : TDef 0
+listByte = TMu "ListByte" [("NilC", T1), ("ConsC", TProd [weakenTDef byte 1 LTEZero, TVar 0])]
+
+charlistNatalphaTriples : TDef 1
```
epost: Could this name be made a bit clearer? I think capitalising it as `charListnatAlpha` would already help a bit.

src/Typedefs.idr
```diff
@@ -156,3 +156,20 @@ mutual
 
 Show (TDef n) where
   show = showTDef
+
+-- Equality -----
```
epost: In keeping with the rest of this file, which is pretty much the main `Typedefs` thing, I propose to align on `==`. This file is pretty central to Typedefs so it seems like a good place for a bit of polish.

src/Backend/ReasonML.idr
```diff
@@ -1,73 +1,135 @@
...
+  RMLVoid  :                                 RMLType
+
+  ||| The `unit` (i.e. singleton) type.
+  RMLUnit  :                                 RMLType
+
+  ||| The tuple type, containing two or more further types.
+  RMLTuple  :         Vect (2 + k) RMLType -> RMLType
+
+  ||| A type variable.
+  RMLVar   :         Name                 -> RMLType
```
epost: Alignment whoopsie on the `->`.

kjekac:
> So this will supersede both #85 and #86 right?

Yes correct.

src/Test.idr
kjekac: Changed it to `tripleByteListnatAlpha`, in keeping with the naming convention where the parametric type comes before the parameters. Good enough?

kjekac: I take the :heart: as an approval. :)

src/Backend/Haskell.idr
```diff
@@ -1,68 +1,127 @@
...
+
+import Data.Vect
+
+%default total
 %access public export
 
-guardPar : String -> String
-guardPar str = if any isSpace $ unpack str then parens str else str
+||| The syntactic structure of Haskell types.
+data HsType : Type where -- TODO could be interesting to index this by e.g. used variable names?
```
fredrikNordvallForsberg: Nice! :+1:

src/Test/Haskell.idr
```diff
@@ -1,79 +1,159 @@
...
 testSuite = spec $ do
 
   describe "Haskell code generation tests:" $ do
 
-    it "bit" $ 
-      generate bit 
-        `shouldBe` "\ntype Bit = Either () ()\n"
+    it "bit" $
+      generate bit
+        `shouldBe` text "type" |++| text "Bit" |++| equals |++| text "Either" |++| parens empty |++| parens empty
```
fredrikNordvallForsberg: A bit nitpicky, but would it make sense to compare the test cases to the actual printed strings (that is,
```
it "bit" $
      (print $ generate bit)
        `shouldBe` "\ntype Bit = Either () ()\n"
```
for instance) so that there is not a bug in the `Doc` rendering?

kjekac: My reasoning for this choice is that whether the tests pass or not shouldn't depend on how we print the `Doc` term. `Doc` needs parameters in order to print (I defined the `print` function here with certain parameters as a convenience), so if one of those is changed, the actual and expected results might differ even though the generated code is essentially the same.

But I agree that the current situation is also pretty clunky, so I don't have a strong preference towards it. Open to arguments and/or alternative suggestions.

src/Backend.idr
```diff
@@ -0,0 +1,102 @@
...
+  generateTyDefs : Env n -> TDef n -> List lang
+
+  ||| Given a type definition in the target language, generate its code.
+  generateCode : lang -> Doc
+
+  ||| Generate a new environment with n free names.
+  freshEnv : (n: Nat) -> Env n
+
+||| Generate code in a specific language for a type definition and all its dependencies.
+||| Needs to be called with the implicit parameter `lang`, as such: `generate {lang=Haskell} td`.
```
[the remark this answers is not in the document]
kjekac: I actually tried to do this, but didn't succeed. I tried again now, with the following code:

```idris
generate : Backend lang => {n: Nat} -> (lang: Type) -> TDef n -> Doc
generate {n} lang td = vsep2 . map (generateCode) . generateTyDefs {lang=lang} (freshEnv {lang=lang} n) $ td
```

Which gives me the error
```
Backend.idr:70:24-108:
   |
70 | generate {n} lang td = vsep2 . map (generateCode) . generateTyDefs {lang=lang} (freshEnv {lang=lang} n) $ td
   |                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
When checking right hand side of generate with expected type
        Doc

When checking an application of function Prelude.Basics..:
        Can't find implementation for Backend lang
```

I understand this as that it isn't inferred that `Backend lang` and `(lang: Type)` refer to the same type. If you know how to solve this I'm all ears, because I agree that the current situation is ugly.

src/Backend/Haskell.idr
kjekac: Are you thinking of the `TODO`? Because that turned out very messy, when I tried. OTOH the rest of my code hadn't stabilized then, so it might be easier now. Seems reasonable to take care of this as a separate issue once this monster has been merged though. :)

src/Backend.idr
```diff
@@ -0,0 +1,102 @@
...
+
+  ||| The names of the type's parameters.
+  params : Vect n Name -- TODO is number of parameters enough?
+
+||| A variable environment. Left=free, Right=bound.
+Env : Nat -> Type
+Env k = Vect k (Either String Decl)
+
+||| Vertically concatenate a list of documents with two newlines (i.e. one empty line) as separator.
+vsep2 : List Doc -> Doc
```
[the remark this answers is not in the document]
kjekac: Yeah very good point, I moved some stuff around between `Backend` and `Backend.Utils` in 40b9fb4b485777bd8ce0db5b076cf219130c01f6.

src/Test/Haskell.idr
fredrikNordvallForsberg: Yes indeed, harmless reformattings of the same code should ideally not matter... Perhaps we could separately test that `ghc` accepts the code? Since layout matters for eg Haskell, it would be good to know that we actually generate valid code. But indeed, this is a minor issue.

src/Backend.idr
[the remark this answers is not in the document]
epost: Good point, though I think addressing this in a separate PR might be good enough. Perhaps we could just make an issue out of this?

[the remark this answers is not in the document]
clayrat: @kjekac this is because you shadow over lang with you later implicit. If you write it as 
```
generate : (lang: Type) -> Backend lang => {n: Nat} -> TDef n -> Doc
generate lang {n} td = vsep2 . map (generateCode) . generateTyDefs {lang} (freshEnv {lang} n) $ td
```
it seems to work.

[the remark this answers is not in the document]
clayrat: Also a minor nitpick of mine is that you can always shorten `{foo=foo}` to just `{foo}` :)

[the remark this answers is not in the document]
kjekac: Thanks Alex! Pushed an update addressing both this and your nitpicks. :)

PR #102, opened 2019-01-15
First version of JSON backend

kjekac:
Issue 92 https://github.com/typedefs/typedefs/issues/92

This does not yet use the same interface as the other two backends, instead we have `NewBackend` and `newGenerate`.


epost:
Any idea what's causing those Travis probs @kjekac ?

src/Backend.idr
```diff
@@ -76,6 +76,14 @@ interface Backend lang where
 generate : (lang: Type) -> Backend lang => {n: Nat} -> TDef n -> Doc
 generate lang {n} td = vsep2 . map (generateCode) . generateTyDefs {lang} (freshEnv {lang} n) $ td
 
+interface NewBackend def type | def where
```
wires: Maybe its nice to add a short doc comment here that describes what `msgType`, `typedefs` and `source` are supposed to do?

kjekac:
> Any idea what's causing those Travis probs @kjekac ?

Import order, but it's fixed now!

src/Backend/JSON.idr
```diff
@@ -0,0 +1,124 @@
...
+                                 , ("required", JArray . toList $ map JString names)
+                                 , ("additionalProperties", JBoolean False)
+                                 , ("properties", JObject . toList $ zip names (map (assert_total makeSubSchema) tds))
+                                 ]
+
+||| Only perform an action if a name is not already present in the state. If the action is performed, the name will be added.
+ifNotPresent : Name -> State (List Name) (List (String, JSON)) -> State (List Name) (List (String, JSON))
+ifNotPresent name gen = do
+    st <- get
+    if name `Prelude.List.elem` st
```
clayrat: I think it's usually enough to disambiguate with the last part of the package, i.e. `List.elem`

src/Backend/Haskell.idr
```diff
@@ -122,6 +122,11 @@ Backend Haskell where
   generateCode        = renderDef
   freshEnv            = freshEnvLC
 
+NewBackend Haskell HsType where
```
clayrat: Is the absence if `NewBackend` instance for ReasonML intentional?

kjekac: Kind of? I'm not even sure why I added the instance for `Haskell` tbh, I think it was mostly to check that `NewBackend` actually is compatible with the old backends. I mean, neither `Backend` nor `NewBackend` will be kept as-is, so I didn't see the point on spending too much time on them. *But* now that you ask, I realize that it's good hygiene that it's good to keep `master` "complete" at all times, so I'll add a `ReasonML` instance.

PR #109, opened 2019-02-27
Application, naming and synchronization of backend interfaces (without specialisation)

kjekac:
### Application and naming

This is a rather large change because it changes the structure of the `TDef` type, which more or less all files work with. Namely, it introduces a constructor `TApp` that represents full application of a `TDef` to a vector of arguments, i.e. naively: `TApp : TDef k -> Vect k (TDef n) -> TDef n`.

Because this naive approach would introduce redundancy (there is literally no semantic difference between `TSum [T1, TVar 0]` and `TApp (TSum [T1, TVar 0]) [TVar 0]`), we restrict the first argument to only _named_ types, which requires us to factor out the naming of types into a separate datatype `TNamed` that is mutually recursive with `TDef`, like so:

```idris
mutual
  ||| Type definition
  ||| @n The number of type variables in the type
  data TDef : (n:Nat) -> Type where
    [...]
    ||| Application of a named type to a vector of arguments.
    TApp  : TNamed k -> Vect k (TDef n)           -> TDef n

  ||| Named type definition.
  ||| @n The number of type variables in the type.
  data TNamed : (n : Nat) -> Type where
    TName : Name -> TDef n -> TNamed n
```

And then basically everything else, except for the changes to the backend interfaces, follows "naturally" from this change.

Two somewhat major things to note:
1. Because names have been eliminated from the `TDef` type, recursive types are now anonymous by default. Code generators are recommended to give special treatment to the structure `TName name (TMu [...])`, but this is not strictly required.
2. The way we used to do type specialisation is believed to be incompatible with these changes, as I explained on RocketChat. For the moment everything relating to this is simply commented out.

### Backend interfaces

The backend interface has been split into three separate interfaces, of which a backend will want to implement two.

Instead of one monolithic interface which deals with both the translation from a `TDef` to an AST and the translation of an AST to text, we have one interface for the first step:

```idris
||| Interface for interpreting type definitions as ASTs.
||| @def  the type representing type definitions.
||| @type the type representing type signatures.
||| @n    the number of variables the interpretor supports in a type definition. (Should always be either `n` or `0`.)
interface ASTGen def type (n : Nat) | def where
  ||| Given a `TDef`, generate its corresponding type signature.
  msgType        : TNamed n -> type

  ||| Generate definitions for a `TNamed` and all its helper definitions.
  generateTyDefs : TNamed n -> List def
```

...and two interfaces for the second step, depending on the concrete syntax of the target language:

```idris
||| Interface for code generators that can generate code for type definitions and
||| type signatures independently of each other.
||| @def  the type representing type definitions.
||| @type the type representing type signatures.
interface CodegenIndep def type | def where
  ||| Generate source code for a type signature.
  typeSource : type -> Doc

  ||| Generate source code for a type definition.
  defSource  : def -> Doc

||| Interface for code generators that need to generate code for type definitions and
||| type signatures at the same time.
||| @def  the type representing type definitions.
||| @type the type representing type signatures.
interface CodegenInterdep def type where
  ||| Generate source code for a type signature and a list of helper definitions.
  sourceCode   : type -> List def -> Doc
```

Of the backends we already have, Haskell and ReasonML implement the first, while JSON implements the second because it doesn't make sense to generate the code for either the "type signature" (i.e. schema for the outermost type) or the "type definitions" (i.e. definition schemas) in isolation, since they need to be put together into one larger top-level schema.


src/Typedefs.idr
```diff
@@ -3,34 +3,129 @@ module Typedefs
...
-  T1    :                                      TDef n
+    ||| Application of a named type to a vector of arguments.
+    TApp  : TNamed k -> Vect k (TDef n)           -> TDef n
 
-  ||| The coproduct type
-  TSum  :         Vect (2 + k) (TDef n)     -> TDef n
+  ||| Named type definition.
+  ||| @n The number of type variables in the type.
+  data TNamed : (n : Nat) -> Type where
+    TName : Name -> TDef n -> TNamed n
```
kjekac: I have the idea that it would be "nice" to avoid the mutual recursion of the name-type and the type-type by letting the name-type be parametric in the type-type, like so:

```idris
||| Named things that are indexed by a `Nat`. (For example `TDef`s.)
||| @thing The type of things to name.
||| @n The 
data Named : (thing : Nat -> Type) -> (n : Nat) -> Type where
  Name : Name -> thing n -> Named thing n

||| Type definition
||| @n The number of type variables in the type
data TDef : (n:Nat) -> Type where
  [...]
  ||| Application of a named type to a vector of arguments.
  TApp  : Named TDef k -> Vect k (TDef n)           -> TDef n
```

It seems to work very well, but except for the "nice feeling" of not having mutually recursive data types, we don't seem to gain very much because none of the mutually recursive functions dealing with these types are easy to turn non-mutually recursive.

src/Typedefs.idr
```diff
@@ -3,34 +3,129 @@ module Typedefs
...
+shiftVars : TDef n -> TDef (S n)
+shiftVars T0          = T0
+shiftVars T1          = T1
+shiftVars (TSum ts)   = assert_total $ TSum $ map shiftVars ts
+shiftVars (TProd ts)  = assert_total $ TProd $ map shiftVars ts
+shiftVars (TVar v)    = TVar $ shift 1 v
+shiftVars (TMu cs)    = assert_total $ TMu $ map (map shiftVars) cs
+shiftVars (TApp f xs) = assert_total $ TApp f $ map shiftVars xs 
+
+||| Get a list of the de Brujin indices that are actually used in a `TDef`.
```
epost:
```suggestion
||| Get a list of the De Bruijn indices that are actually used in a `TDef`.
```

src/Backend.idr
```diff
@@ -12,86 +12,49 @@ import Text.PrettyPrint.WL
...
-
-||| Filter out the entries in an `Env` that is referred to by a `TDef`.
-getUsedVars : Env n -> (td: TDef n) -> Env (length (getUsedIndices td))
-getUsedVars e td = map (flip index e) (fromList $ getUsedIndices td)
+||| Interface for interpreting type definitions as ASTs.
+||| @def  the type representing type definitions.
+||| @type the type representing type signatures.
+||| @n    the number of variables the interpretor supports in a type definition. (Should always be either `n` or `0`.)
+interface ASTGen def type (n : Nat) | def where
+  ||| Given a `TDef`, generate its corresponding type signature.
```
epost: Shouldn't this be
```suggestion
  ||| Given a `TNamed`, generate its corresponding type signature.
```
?

src/Backend.idr
```diff
@@ -12,86 +12,49 @@ import Text.PrettyPrint.WL
...
-  ||| Given a `TDef` and a matching environment, generate a list of type definitions
-  ||| in the target language.
-  generateTyDefs : Env n -> TDef n -> List lang
+  ||| Generate definitions for a `TNamed` and all its helper definitions.
+  generateTyDefs : TNamed n -> List def
 
-  ||| Given a type definition in the target language, generate its code.
-  generateCode : lang -> Doc
+||| Interface for code generators that can generate code for type definitions and
+||| type signatures independently of each other.
```
fredrikNordvallForsberg: It might be nice to list some examples here, e.g. from the PR description 
"Haskell and ReasonML".

src/Backend.idr
```diff
@@ -12,86 +12,49 @@ import Text.PrettyPrint.WL
...
-  typedefs : TDef 0 -> List def
+||| Use the given backend to generate code for a type signature.
+generateType : (def: Type) -> ASTGen def type n => CodegenIndep def type => TNamed n -> Doc
+generateType def tn = typeSource {def} (msgType {def} tn)
 
-  ||| Given a type signature and a list of helper definitions which it uses,
-  ||| generate source code.
-  source   : type -> List def -> Doc
+||| Interface for code generators that need to generate code for type definitions and
+||| type signatures at the same time.
```
fredrikNordvallForsberg: Similarly, it might be nice to list some examples also here, e.g. from the PR description 
"JSON".

src/Backend/Utils.idr
```diff
@@ -1,28 +1,83 @@
...
+    flattenMu _     T1          = T1
+    flattenMu names (TSum ts)   = assert_total $ TSum $ map (flattenMu names) ts
+    flattenMu names (TProd ts)  = assert_total $ TProd $ map (flattenMu names) ts
+    flattenMu names td@(TMu cs) = def $ flattenMu' names $ TName (nameMu cs) td
+    flattenMu names (TApp f xs) = assert_total $ TApp f (map (flattenMu names) xs)
+
+    flattenMu' : Vect (S n) Name -> TNamed (S n) -> TNamed n
+    flattenMu' names (TName name body) = case body of
+     TMu cases => assert_total $ TName name $ TMu $ map (map (flattenMu (name :: names))) cases
+     _         => assert_total $ TName name $ flattenMu names body
```
fredrikNordvallForsberg: This catchall will never be reached, right? The function `flattenMu'` is only called on line 77, with exactly an argument of
```
TName (nameMu cs) (TMu xs)
```
so also taking the `def` call there into account, couldn't this be inlined and line 77 be replaced by
```
flattenMu names (TMu cs) = TMu $ map (map (flattenMu ((nameMu cs) :: names))) cs

```
?

src/Backend/Utils.idr
```diff
@@ -1,28 +1,83 @@
...
+vsep2 = vsep . punctuate line
+
+||| Produce a name for an anonymous `TMu` by simply concatenating its constructors.
+nameMu : Vect n (Name, TDef k) -> Name
+nameMu = concatMap (uppercase . fst)
+
+||| "Flatten" all `TVar` references to `TMu`s into `TName`s named after the corresponding `TMu`, referencing `T0`s.
+||| This is strictly meant to remove any `TVar`s from an AST.
+||| The variable in the argument `TDef 1` is expected to reference the supplied name.
+flattenMus : Name -> TDef 1 -> TDef 0 -- TODO this should take a proof that all variables refer to a TMu.
```
fredrikNordvallForsberg: Could this perhaps be achieved using `ap` instead?

src/Backend/Utils.idr
```diff
@@ -1,28 +1,83 @@
...
+flattenMus muName = flattenMu [muName]
+  where
+  mutual
+    -- DO NOT simply lift this function out to the top-level.
+    -- Its behavior is dependent on the type of `makeDefs'`.
+    -- (Specifically: all TVars must refer to a TMu, not to any free variables.)
+    flattenMu : Vect (S n) Name -> TDef (S n) -> TDef n
+    flattenMu names (TVar v)    = wrap $ TName (index v names) T0
```
fredrikNordvallForsberg: I'm a bit nervous of this `T0`. I guess the body is never used?

src/Typedefs.idr
```diff
@@ -3,34 +3,129 @@ module Typedefs
...
+
+square : String -> String
+square "" = ""
+square s = "[" ++ s ++ "]"
 
-  ||| Mu
-  TMu   : Name -> Vect k (Name, TDef (S n)) -> TDef n
+||| Generate the canonical name of a closed type.
+makeName : TDef 0 -> Name
+makeName T0          = "emptyType"
```
fredrikNordvallForsberg:
```suggestion
makeName T0          = "void"
```

src/Typedefs.idr
```diff
@@ -3,34 +3,129 @@ module Typedefs
...
+makeName T1          = "singletonType"
```
fredrikNordvallForsberg:
```suggestion
makeName T1          = "unit"
```

src/Backend/Utils.idr
kjekac: I completely agree with you! I dislike this solution. And no, it shouldn't be, because the name that we give it is that of a `TMu` further out in the tree. But when we (err.. you) do specialisation in a nice way, this is a place where that machinery should probably be used.

src/Backend/Utils.idr
kjekac: They might be overlapping, I wrote them at different points in both time and problem-space, so I might have missed an opportunity to implement one in terms of the other. But their functionalities are distinct. The "problem" is that even inside a `TDef 0` generated by `ap`, you might have recursive variables, so in even in backends explicitly state that they don't handle variables, you might encounter a `TDef 1` and need to reduce it to a `TDef 0`.

But using `ap` in the implementation of `flattenMu` might be feasible, I'll have a look soon.

src/Backend/Utils.idr
kjekac: True, thank you! Fixed and pushed.

src/Backend/Utils.idr
kjekac: Upon further inspection: No, I don't see the possibility of implementing one in terms of the other, but if I'm missing something, please show me what. I don't like this function either.

kjekac:
@fredrikNordvallForsberg I really appreciate you looking at `flattenMus`, I don't like that function personally. FWIW, @andrevidela had the idea that it should take a proof that all variables actually refer to a mu. [I implemented the proof type and a few functions around it today](https://github.com/typedefs/typedefs/compare/application-without-specialisation...free-vars-proof) but things actually get a bit strange when adapting `flattenMus` to this. I won't have the time to figure that part out, but I'll leave the branch there if anyone wants to continue the investigations.

src/Backend.idr
```diff
@@ -12,86 +12,49 @@ import Text.PrettyPrint.WL
...
 ||| Use the given backend to generate code for a type definition and all its dependencies.
-generate : (lang: Type) -> Backend lang => {n: Nat} -> TDef n -> Doc
-generate lang {n} td = vsep2 . map (generateCode) . generateTyDefs {lang} (freshEnv {lang} n) $ td
-
-||| Interface for codegens which distinguish between the message type itself and
-||| its helper definitions. Currently doesn't support typedefs with free variables. 
-interface NewBackend def type | def where
-  ||| Given a `TDef`, generate its corresponding type signature.
-  msgType  : TDef 0 -> type
+generateDefs : (def: Type) -> ASTGen def type n => CodegenIndep def type => TNamed n -> Doc
```
clayrat: Constraints can be uncurried to `(ASTGen def type n, CodegenIndep def type) =>`, probably a bit more structured this way?

src/Backend/Haskell.idr
```diff
@@ -78,55 +83,56 @@ renderDef (ADT     decl cases) = text "data" |++| renderDecl decl
...
+makeType' : Env n -> TNamed n -> HsType
+makeType' e (TName name body) = HsParam name . map (either HsVar hsParam) $ getUsedVars e body
+
+mutual
+  ||| Generate all the Haskell type definitions that a `TDef` depends on.
+  makeDefs : TDef n -> State (List Name) (List Haskell)
+  makeDefs    T0          = pure []
+  makeDefs    T1          = pure []
+  makeDefs    (TProd xs)  = map concat $ traverse (assert_total makeDefs) xs
+  makeDefs    (TSum xs)   = map concat $ traverse (assert_total makeDefs) xs
```
clayrat: `concat <$> traverse (assert_total makeDefs) xs` to make it consistent with `TApp` case?

src/Typedefs.idr
```diff
@@ -3,34 +3,129 @@ module Typedefs
...
-  ||| The unit type
-  T1    :                                      TDef n
+    ||| Application of a named type to a vector of arguments.
+    TApp  : TNamed k -> Vect k (TDef n)           -> TDef n
 
-  ||| The coproduct type
-  TSum  :         Vect (2 + k) (TDef n)     -> TDef n
+  ||| Named type definition.
+  ||| @n The number of type variables in the type.
+  data TNamed : (n : Nat) -> Type where
```
clayrat: This looks like it can be made a record, packaging `name` and `def` getters inside.

PR #114, opened 2019-03-15
Code style guide

andrevidela:
I've added the bare minimum for cosmetics when working with Idris. Feel free to chime in with ideas.

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -67,7 +67,7 @@ fn = case True of
 
 Case analysis is indented with 5 spaces
 
-## ASCII-art
+## Alignment ASCII-art
 
 Sometimes, syntax in idris can become very verbose in places where 
 understanding the structure of the program is key.
@@ -76,7 +76,7 @@ understanding the structure of the program is key.
 postulate hardToRead : (N (N a b) c) = (N a (N b c))
 ```
 
-In this case it might be relevant to indent the file as follows
+In this case it might be relevant to align the code as follows
 
 ```
 easierToRead: (N (N a b) (  c  )) =
@@ -84,8 +84,8 @@ easierToRead: (N (N a b) (  c  )) =
 ```
 
 As we can see, the _structure_ of the equality is much easier to see (a form of
-associativity here). Since this is ascii art, there can be a lot of equaly
-valid ways of indenting this expression. Here are some examples
+associativity here). Since this is ascii art, there are multiple equally
+valid ways of aligning this expression. Here are some examples:
 
 1.
 
@@ -104,7 +104,7 @@ easierToRead: (N (N a    b) c) =
               (N    a (N b  c))
 ```
 
-Using operators one could write:
+Using operators, one could write:
 
 4.
 ```
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -1,16 +1,16 @@
 
 # Typedefs styleguide
 
-This styleguide aims to answer questions relevants to the cosmetics related to 
+This styleguide aims to answer questions relevants to the cosmetics related to
 writing Idris programs.
 
-Since idris is a relatively young language, there is no general convention 
-about its style. This document should serve as reference for current and future 
+Since idris is a relatively young language, there is no general convention
+about its style. This document should serve as reference for current and future
 users of the typedefs codebase.
 
 ## Line length: 120
 
-The maximum line length is 120 characters. In today's wide screen monitors 
+The maximum line length is 120 characters. In today's wide screen monitors
 there is no reason to limit ourselves to 80 characters, yet, having a maximum
 limit helps avoiding degenerate situations.
 
@@ -22,7 +22,7 @@ Haskell. In order to allow this level of visual customisation and ensure
 consistency across editors, spaces are used.
 
 The language does not benefit from more than 2 space of indentation in most
-cases. Thought special constructions lend themselve to more spaces for 
+cases. Thought special constructions lend themselve to more spaces for
 indentation.
 
 ### Indentation: where clauses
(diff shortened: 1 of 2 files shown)


STYLEGUIDE.md
```diff
@@ -0,0 +1,133 @@
...
+        g = "hello"
+```
+
+Where clauses are aligned with 6 spaces
+
+### Indentation: Let declarations
+
+```idris
+fn : String
+fn = let greeting = "hello" in
```
clayrat: What about multiclause let's, i.e., 
```let foo = 1
    bar = 2 
  in
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,133 @@
...
+```
+
+Let declarations are indented with 4 spaces
+
+### Indentation: rewrites
+
+```idris
+fn : String
+fn = rewrite proof in
+             1234
```
clayrat: Won't this lead to a "ladder" when you have successive rewrites?

STYLEGUIDE.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Indentation: where clauses
+
+``` idris
+fn : String
+fn = g
+  where g : String
+        g = "hello"
+```
+
```
epost: Currently, one of the styles I see in the codebase (e.g. [here](https://github.com/typedefs/typedefs/blob/0050cdcf1c19ce9bfc53ac35dcad6512be611e6b/src/Parse.idr)) is this:
```idris
fn : String
fn = g
  where
  g : String
  g = "hello"
```
Marco and myself had a discussion on how to do this in the AQL/CQL Haskell codebase, also considering the fact that ~~PureScript requires an indent after where~~. We settled on a 2-space indent after `where`, like this:
```idris
fn : String
fn = g
  where
    g : String
    g = "hello"
```
It may make sense to do the same for Typedefs in order to keep things memorable. (I guess we'd have to indent other constructs similarly.) This has the added benefit of consuming less horizontal space. I am aware that Idris codebases are slightly different beasts, so there may be issues with this, not sure.

STYLEGUIDE.md
```diff
@@ -0,0 +1,133 @@
...
+Idris makes uses of some whitespace sensitive construction (typically the `do`
+notation), and also inherits from a culture of ascii-art code from Agda and
+Haskell. In order to allow this level of visual customisation and ensure
+consistency across editors, spaces are used.
+
+The language does not benefit from more than 2 space of indentation in most
+cases. Thought special constructions lend themselve to more spaces for 
+indentation.
+
```
epost: Following the Halogen-inspired style we use elsewhere, I propose the following for type sigs:

---
### Indentation: type signatures

```idris
fn
   : A
  -> B
  -> c
fn x = 42
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,133 @@
...
+```idris
+fn : String
+fn = case True of
+          True => "Hello"
+          False => "Goodbye"
+```
+
+Case analysis is indented with 5 spaces
+
+## ASCII-art
```
epost:
```suggestion
## Alignment ASCII-art
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,133 @@
...
+
+Sometimes, syntax in idris can become very verbose in places where 
+understanding the structure of the program is key.
+
+```
+postulate hardToRead : (N (N a b) c) = (N a (N b c))
+```
+
+In this case it might be relevant to indent the file as follows
```
epost:
```suggestion
In this case it might be relevant to align the code as follows
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,133 @@
...
+
+```
+easierToRead: (N (N a b) (  c  )) =
+              (N (  a  ) (N b c))
+```
+
+As we can see, the _structure_ of the equality is much easier to see (a form of
+associativity here). Since this is ascii art, there can be a lot of equaly
+valid ways of indenting this expression. Here are some examples
```
epost:
```suggestion
valid ways of aligning this expression. Here are some examples:
```

STYLEGUIDE.md
andrevidela: There is no issue with this, it makes just as much sense. I'll change the description accordingly

STYLEGUIDE.md
andrevidela: I had this in mind

```
let foo = 1
    bar = 2 in
    expr
```

What do you think?

STYLEGUIDE.md
andrevidela: That is true. Do you suggest we have

```
fn = rewrite proof1 in
     rewrite proof2 in
             1234
```
?

STYLEGUIDE.md
fredrikNordvallForsberg: If you want to align `expr` with the bound variables `foo` and `bar`, why not move the `in` to the next line?
```idris
let foo = 1
    bar = 2
in  expr
```

STYLEGUIDE.md
fredrikNordvallForsberg: Personally, I think this looks horrible, but I could live with it. (Some passive aggression there for you.) Even worse, the example doesn't typecheck!

epost: @fredrikNordvallForsberg Thank you for your insightful and introspective feedback! I hope you will find the following amendments soothing:
```idris
-- short type sig, so no need for nasty newlineses
whatever : Int -> Int
whatever x = 42

-- but then, look at all the pretty alignment here!
numberOfTheBeast
   : AnInordinatelyLongTypeInTheHopesOfAppeasingFredrik
  -> YetAnotherExcruciatinglyLongType_ThisBetterWork
  -> Int
numberOfTheBeast x y =
  666
```

marcosh: I'm currently writing multiline function signatures as

```
fn :
     a : Type
  -> b : Type
  -> Type
```

I guess that having a line with ` : a : Type` could be not that clear at first sight

STYLEGUIDE.md
marcosh: I agree with 2 spaces indentation after `where`

STYLEGUIDE.md
marcosh: I like @fredrikNordvallForsberg proposal, or otherwise

```
let
  foo = 1
  bar = 2
in
  expr
```

but it starts to be quite long

STYLEGUIDE.md
```diff
@@ -0,0 +1,133 @@
...
+Rewrite rules are indented with 8 spaces
+
+### Indentation: Case
+
+```idris
+fn : String
+fn = case True of
+          True => "Hello"
+          False => "Goodbye"
+```
```
marcosh: why not simple 2 spaces indentation?

```
case foo of
  True  => "Hello"
  False => "Goodbye"
```

Moreover, it would be nice to keep `=>` aligned

STYLEGUIDE.md
andrevidela: @fredrikNordvallForsberg "If you want to align expr with the bound variables foo and bar, why not move the in to the next line?" 

My argument against it would be that in no other parts of the language or styleguide do we encourage using two spaces between two syntactic terms.

@marcosh I like your proposition better but I agree that it's getting very long. 

If we agree on that indentation

```idris
fn = let a = b in 
         expr
```

(That is, `expr` should be indented with 4 spaces) Then I think it makes sense to stay consistent with the 4 spaces rules and multiple lines indented like so:

```idris
fn = let a = b
         c = d in
         expr
```

STYLEGUIDE.md
andrevidela: @marcosh I think your suggestion makes sense. Moreover the style stays consistent when you introduce type classes

```idris
fn : (myTypeClasss a)
  => a : Type
  -> b : Type
  -> Type
```

I'll add two paragraphs about function signatures and interface declarations. As well as argument grouping

STYLEGUIDE.md
andrevidela: I've updated the `let` paragraph with 2 spaces instead of 4. Tell me what you think

STYLEGUIDE.md
andrevidela: I've added a section about that, please tell me what you thinkg

andrevidela:
@clayrat @fredrikNordvallForsberg @epost @marcosh 
I've updated some paragraphs with your suggestions and added more, could you please take a look?

STYLEGUIDE.md
```diff
@@ -0,0 +1,256 @@
...
+Let declarations are indented with 2 spaces.
+
+Additional declarations are indented with 4 spaces while the expression is indented with 2 spaces
+
+```idris
+fn : String
+fn = let greeting = "hello"
+         whom = "world" in
+       greeting
+```
```
marcosh: this doesn't look that nice, from my point of view, `greeting` being not aligned with everything. I'm afraid it won't be that readable in the middle of some complex code: Personally I second the @fredrikNordvallForsberg proposal

STYLEGUIDE.md
```diff
@@ -0,0 +1,256 @@
...
+fn = rewrite proof in
+       1234
+```
+
+Rewrite rules are indented with 2 spaces. Multiple rewrite rules in a row are not indented.
+
+```idris
+fn : String
+fn = rewrite proof in
+     rewrite prrof2 in
```
marcosh: typo on `prrof2`

STYLEGUIDE.md
```diff
@@ -0,0 +1,256 @@
...
+definitions, it is expected to break it up as follows:
+
+```idris
+veryLongFunctionBecauseOfInterfaces :
+   ( Interface1 a
+   , Interface2 b
+   , Interface3 c )
+  => (a, b, c : Type)
+  -> Type
+```
```
marcosh: nice, I really like this!

STYLEGUIDE.md
andrevidela: fixed

STYLEGUIDE.md
andrevidela: I think I've come around to you and @fredrikNordvallForsberg but I'm still not convinced about the two spaces between `in` and the expression. What do you think of

```idris
let foo = 1
    bar = 2
 in expr
```

?

marcosh: looks nice, too!

Issue #49, opened 2018-10-01
JS API should expose top-level methods


epost:
## Outline

The idea is that [ParserJS.idr](https://github.com/typedefs/typedefs/blob/98bf628abb94cc703d566ece2310c2e659a0fcda/parser.js/ParserJS.idr) call the various [Backend](https://github.com/typedefs/typedefs/tree/98bf628abb94cc703d566ece2310c2e659a0fcda/src/Backend)s instead of just the Haskell one.

- [x] expose Haskell backend
- [x] expose ReasonML backend
- [x] expose JSON backend
- [x] expose type parser
- [x] remove `main` function
- [x] add a `String` switch to select the desired backend

## Older version of this issue (plz update!)

This is a probably somewhat outdated TODO:

- [ ] expose `showTDef`
- [ ] ~~expose `Backend.Haskell`: `generate`, `generateType`~~
- [ ] ~~expose `TermCodec` (`serialize` and `serializeMu`?)~~

Re invocation of the Haskell backend, @clayrat writes

> It's generally `generate <typedef>` or `generateType <typedef>`
> depends on whether you're making type synonyms or actual new data
> I'll try to add those tests later today

See https://github.com/typedefs/typedefs/issues/48.

Alex also mentioned https://github.com/idris-lang/Idris-dev/pull/4073, which should allow us to create interfaces for Node (and, after some browserification, hopefully also for the browser).

andrevidela:
copy API from Haskell

PR #115, opened 2019-03-20
expose/reexport backends and term codec

epost:
@clayrat As far as I can see this PR is good to go. We do want the backend switch string Jelle mentioned yesterday, but we could do that in a separate PR.


epost:
Oh drat, the build failed. :(

clayrat:
Hm, somehow I missed one import. I guess I'll re do the backend function with a switch in this PR anyway, so we won't need that.

clayrat:
Closes #49

epost:
Fabuloso! Thanks @clayrat! 🌈

PR #116, opened 2019-03-24
WIP typedefs as a category

andrevidela:
This PR implements the category of typedefs. Objects are Typedefs, morphisms are function between typedefs. 

The category is parameterized over the number of free variables in the typedef. It might be that we can relax this constraint but I have not tried.

My first attempt was to convert typedefs into idris native types. But I don't think there is any way to define recursive definition as-is. It might be possible using laziness but this seemed quite unintuitive to me.

### DIFF
--- a/src/Typedefs/Category.idr
+++ b/src/Typedefs/Category.idr
@@ -13,8 +13,7 @@ import Interfaces.Verified
 
 %default total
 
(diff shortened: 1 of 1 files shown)


andrevidela:
This has been implemented in some from elsewhere so closing it for now

PR #119, opened 2019-04-03
Replace explanation in README by a link to typedefs.com

epost:


epost:
Branched off the wrong branch :(

PR #127, opened 2019-04-06
fix elba configuration

marcosh:


andrevidela:
I was about to open the same pull request after installing with elba. Looks good

PR #131, opened 2019-04-16
fix multi-line comments

### LINKED ISSUE #118: Parser fails on multiple consecutive line comments
This works:
```
; this works nicely
(+ 1 1)
```
This doesn't:
```
; this does
; not work
(+ 1 1)
```
Nor does this:
```
; adding an empty line also does not work

(+ 1 1)
```

While we're on it, I would also suggest to name some of the functions introduced [in this PR](github.com/typedefs/typedefs/pull/113/files#diff-3992aafdffb714eca9c007eb3aed29a9R59) something with `lineComment` in it, and to distinguish between that and `lineComments` (plural).

andrevidela:
fix #118 
@clayrat  could you take a look?


clayrat:
@andrevidela I've simplified the parsers, and the tests still pass, what do you think?

andrevidela:
Looks good!

PR #140, opened 2019-05-27
Command line options

andrevidela:
fix #140 by adding `--help` and options for input and output files to typedef_parser

## caveats

This uses https://github.com/HuwCampbell/optparse-idris which has a mistake in how it runs the parser. I've used modified version of the library and will submit a pull request later.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -42,6 +42,22 @@
 }:
 
 let
+  config = {
+    packageOverrides = pkgs: {
+      idrisPackages = pkgs.idrisPackages.override {
+        # We override the 'idrisPackages' packaget set
+        # This means, that all these package definitions will 'see' eachother
+        #  e.g. if typedefs-parser depends on typedefs, callPackage will automatically inject it
+        # Add new idris packages to this list 
+        overrides = idrisPackagesNew: idrisPackagesOld: {
+          typedefs = idrisPackagesNew.callPackage ./typedefs.nix {};
+          typedefs-parser = idrisPackagesNew.callPackage ./typedefs-parser.nix {};
+          typedefs-parser-js = idrisPackagesNew.callPackage ./typedefs-parser-js.nix {};
+          typedefs-examples = idrisPackagesNew.callPackage ./typedefs-examples.nix {};
+        };
+      };
+    };
+  };
   syspkgs = import <nixpkgs> { };
   pinpkgs = syspkgs.fetchFromGitHub {
     owner = "NixOS";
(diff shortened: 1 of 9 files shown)


parser/Parser.idr
```diff
@@ -4,11 +4,129 @@ import Backend
...
+    example:
+      typedefs_parser -s bool.tdef
+
+  --inline, -i : input as a string inline
+    example:
+      typedefs_parser -i "(name bool (+ 1 1))" --stdout
+
+  --dest, -d : path to the destination file
+    example:
+      typedefs_parser -i bool.tdef project/typedefs/types.hs
```
clayrat: Shouldn't this be `-s` instead of `-i`? Same for the example below. Also `-d` seems to be missing.

clayrat:
Look cool, make let's put the cmdline parsing stuff in a new module e.g. `Cmdline`?

parser/Parser.idr
andrevidela: You're right, it should either be 
`-i "(+ 1 1)"` or `-s bool.tdef`

andrevidela:
excellent idea @clayrat I'll also submit a PR to `optparse` to add the `flag'`

andrevidela:
I've added the WIP tag since I still need to fix the nix build, the tests and move the command line stuff into its own file

parser/Parser.idr
andrevidela: should be fixed

andrevidela:
Still need to fix nix, I've submitted a PR to [optparse](https://github.com/HuwCampbell/optparse-idris/pull/8) to include `flag'` and fix their `runParserFully` function

default.nix
```diff
@@ -51,19 +68,20 @@ let
...
   };
-  usepkgs = if null == pkgs then
-             import pinpkgs {}
-            else
-              if 0 == pkgs then
-                import <nixpkgs> { }
-              else
-                import pkgs {};
+  usepkgs = (
+    if null == pkgs
```
clayrat: Is it ok to remove indentation here?

parser/Parser.idr
```diff
@@ -3,12 +3,40 @@ import Parse
...
+writeOutput : OutputFile -> String -> IO ()
+writeOutput StdOutput tdef = putStrLn tdef
+writeOutput (FileOutput path) tdef = do Right () <- writeFile path tdef
+                                          | Left error => putStrLn ("File write error: " ++ show path)
+                                        putStrLn ("Generated typedef at " ++ path)
+getInput : InputFile -> IO (Either FileError String)
+getInput (StringInput x) = pure (Right x)
+getInput (FileInput x) = readFile x
+
+generateTDef : String -> Maybe String
```
clayrat: maybe call it `parseAndGenerateTDef`?

PR #141, opened 2019-06-04
Error reporting in parser

clayrat:
This requires https://github.com/gallais/idris-tparsec/pull/23 to be merged first.


clayrat:
@epost @andrevidela https://github.com/gallais/idris-tparsec/pull/23 is now merged, it should be ready now.

src/Parse.idr
```diff
@@ -44,23 +42,34 @@ fromVMax {m} vm = go lteRefl vm
...
+getResult : Result Error (Maybe a) -> Result Error a
+getResult res = res >>= (Result.fromMaybe RunError)
 
 comment : (Alternative mn, Monad mn, Subset Char (Tok p), Eq (Tok p), Inspect (Toks p) (Tok p)) =>
            All (Parser mn p ())
-comment = cmap () $ and (char ';') (roptand (nelist $ notChar '\n') (char '\n'))
+comment = cmap () $ and (char ';') (roptand (nelist $ anyCharBut '\n') (char '\n'))
 
-spacesOrComments : (Alternative mn, Monad mn, Subset Char (Tok p), Inspect (Toks p) (Tok p), Eq (Tok p)) =>
+spacesOrComments : (Alternative mn, Monad mn, Subset Char (Tok p), Inspect (Toks p) (Tok p), Eq (Tok p)) =>
```
andrevidela: Extra space?

src/Parse.idr
```diff
@@ -83,12 +93,13 @@ tdef =
               )
               (parens (and (guardM (\(mp, nam) => pushName nam <$> lookup nam mp) $ mand (lift get) alphas)
                            (map {a=Parser' _} (nelist . ignoreSpaces) rec)))
-        , cmap (Z ** T0) $ string "0"
-        , cmap (Z ** T1) $ string "1"
+        ,
```
andrevidela: Extra line break too

src/Parse.idr
```diff
@@ -83,12 +93,13 @@ tdef =
...
-        , cmap (Z ** T0) $ string "0"
-        , cmap (Z ** T1) $ string "1"
+        , 
+          cmap (Z ** T0) $ char '0'
+        , cmap (Z ** T1) $ char '1'
         , nary rec '*' TProd
         , nary rec '+' TSum
         , map (\n => (S n ** TVar $ last {n})) $
-            parens (rand (ignoreSpaces (string "var")) (ignoreSpaces decimalNat))
+            parens $ rand (ignoreSpaces $ string "var") (ignoreSpaces $ commit $ decimalNat)
```
andrevidela: What does `commit` do?

clayrat: Internally it turns soft fails into hard ones, limiting backtracking. This is essentially what allows error positions to be more or less specific, otherwise you'll just report the root of the whole offending tree.

clayrat:
Yeah it is sort of a leak, basically this comes from the fact we're using `show` to print `Result`s, which is defined like this: https://github.com/gallais/idris-tparsec/blob/master/src/TParsec/Result.idr#L33

I'm not sure what's the best way here, should we just ditch `show`and construct the message manually, or is there some more clever way?

src/Parse.idr
andrevidela: Ah I see, thanks :)

andrevidela:
Since `Result` is a public export we can just write our own `Result -> String` and that would be enough

andrevidela:
Actually the error in CI is a problem. We should fix it before merging it upstream. I happens on my computer too and is a result of the new verison of `tparsec` I just tried compiling with the old one and the problem does not appear.

andrevidela:
Actually I got the versions wrong, everything works fine.

clayrat:
Fixed the syntax and switched tests to work against `Result` instead of raw strings, this required another PR to tparsec which is already merged. Regrading the printer thing, I've realized we already have `parseTDefThenStrFun`, and using it for printing will avoid exposing the hard/soft fail details.

clayrat:
@andrevidela @epost should we merge this or do you have more comments?

andrevidela:
@clayrat we should but after fixing the tparsec dependency obviously

clayrat:
@epost @andrevidela I've sorta fixed the build by removing postinstall commands and adding direct dependencies on tparsec in derived projects, now should we carry these changes onto master?

andrevidela:
@clayrat What does removing the post install command do? doesn't it mean that the binaries and libraries won't be installed?

About the rebase: I think you can keep the JS api that's on `master` and return `Maybe String`, I'll do the conveversion to `Either Error String` after this is merged.

PR #142, opened 2019-06-08
WIP - Expose terms and type signatures in JS

### LINKED ISSUE #130: Expose term codec in JS API

andrevidela:
fix #130 

I've split up the API between generating types and generating seralizers for terms that inhabit those types. I would like to clean up the typeclasses that expose this api to have a clear distiniction between generating the type signatures and generating the serializers. What do you think?

@epost could you take a look and try it?

I've also simplified the API so that we only ever interact with strings an no sigma types.

### DIFF
--- a/parser.js/ParserJS.idr
+++ b/parser.js/ParserJS.idr
@@ -11,18 +11,26 @@ import Backend.ReasonML
 
 generateCode : String -> (n ** TNamed n) -> String
 generateCode "haskell"  (n  **tn) = toString $ generateDefs Haskell tn
-generateCode "reasonml" (n  **tn) = toString $ generateDefs ReasonML tn
-generateCode "json"     (Z  **tn) = toString $ generate JSONDef tn
-generateCode "json"     (S _**tn) = "<error : can't generate JSON schema for open typedef>"
-generateCode _          _         = "<error : unknown backend>"
+generateCode _          _         = "<error : unsupported backend>"
 
 -- re-exports
 parseType : String -> Maybe (n : Nat ** TNamed n)
 parseType = parseTNamed 
 
+generateSerializers : String -> String -> Maybe String
+generateSerializers backend tdef = map (generateCode backend) (parseType tdef)
+
+genType : String -> (n ** TNamed n) -> String
+genType "reasonml" (n   ** tn) = toString $ generateDefs ReasonML tn
+genType "json"     (Z   ** tn) = toString $ generate JSONDef tn
+genType "json"     (S _ ** tn) = "<error : cannot generte JSON schema for open typedefs>"
+genType _          _           = "<error : unsupported backend>"
+
+generateTypeSignature : String -> String -> Maybe String
+generateTypeSignature backend tdef = map (genType backend) (parseType tdef)
+
 lib : FFI_Export FFI_JS "" []
-lib = Data (n ** TNamed n) "TNamedN" $
-      Data (Maybe (n ** TNamed n)) "MaybeTNamedN" $
-      Fun parseType "parseType" $
-      Fun generateCode "generateCode" $
+lib = Data (Maybe String) "MaybeString" $
+      Fun generateSerializers "generateSerializers" $
+      Fun generateTypeSignature "generateTypeSignature" $
       End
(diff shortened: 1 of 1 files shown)


andrevidela:
@clayrat Can you take a look at this? I've tried generating the `.js` code using the `typedefs-parser-js.ipkg` and it doesn't seem to work _anymore_ (it used to when I submitted the PR). Using the following program I see that the exported object is empty 

```javascript
var typedefs = require('./typedefs_parser.js')

console.log(typedefs)
```

Is it empty for you as well?

clayrat:
Doesn't seems so:
```
$ node test.js 
{ generateSerializers: [Function: Main__generateSerializers],
  generateTypeSignature: [Function: Main__generateTypeSignature] }
```

parser.js/ParserJS.idr
```diff
@@ -11,18 +11,26 @@ import Backend.ReasonML
...
 generateCode : String -> (n ** TNamed n) -> String
 generateCode "haskell"  (n  **tn) = toString $ generateDefs Haskell tn
-generateCode "reasonml" (n  **tn) = toString $ generateDefs ReasonML tn
-generateCode "json"     (Z  **tn) = toString $ generate JSONDef tn
-generateCode "json"     (S _**tn) = "<error : can't generate JSON schema for open typedef>"
-generateCode _          _         = "<error : unknown backend>"
+generateCode _          _         = "<error : unsupported backend>"
 
 -- re-exports
 parseType : String -> Maybe (n : Nat ** TNamed n)
```
epost: We no longer export `parseType`, so this comment isn't accurate anymore, right?

parser.js/ParserJS.idr
```diff
@@ -11,18 +11,26 @@ import Backend.ReasonML
...
+generateSerializers : String -> String -> Maybe String
+generateSerializers backend tdef = map (generateCode backend) (parseType tdef)
+
+genType : String -> (n ** TNamed n) -> String
+genType "reasonml" (n   ** tn) = toString $ generateDefs ReasonML tn
+genType "json"     (Z   ** tn) = toString $ generate JSONDef tn
+genType "json"     (S _ ** tn) = "<error : cannot generte JSON schema for open typedefs>"
+genType _          _           = "<error : unsupported backend>"
+
+generateTypeSignature : String -> String -> Maybe String
```
epost: 1) The naming difference between `genType` and `generateTypeSignature` seems a bit ad hoc; perhaps `generateTypeSignature'` would be 'better', instead of `genType`?
2) Re the naming of `generateSerializers` vs `generateTypeSignature`: The first seems to me to be a serialiser of terms, and the second a serialiser of types.
3) Also, isn't having `Signature` in there a bit superfluous?

parser.js/ParserJS.idr
```diff
@@ -11,18 +11,26 @@ import Backend.ReasonML
...
 parseType : String -> Maybe (n : Nat ** TNamed n)
 parseType = parseTNamed 
 
+generateSerializers : String -> String -> Maybe String
+generateSerializers backend tdef = map (generateCode backend) (parseType tdef)
+
+genType : String -> (n ** TNamed n) -> String
+genType "reasonml" (n   ** tn) = toString $ generateDefs ReasonML tn
+genType "json"     (Z   ** tn) = toString $ generate JSONDef tn
+genType "json"     (S _ ** tn) = "<error : cannot generte JSON schema for open typedefs>"
```
epost:
```suggestion
genType "json"     (S _ ** tn) = "<error : cannot generate JSON schema for open typedefs>"
```

parser.js/ParserJS.idr
andrevidela: This is true. I'll remove it :)

parser.js/ParserJS.idr
andrevidela: 1: `genType` is private and only used as a helper for `generateTypeSignature` maybe I can move it as à where clause?

2-3: would renaming the second one as `generateType` or `generateTypeDefinition` or `generateDatatypes` be more helpful?

epost: 1) `where` sounds good!

2) `generateType` is lovely, I think. 👍On top of that, don't you think having the word 'term' in the name `generateSerializers` somewhere would be good? Perhaps `generateTermSerializers`? It is a bit long, but for reasons of searchibility, it may be worth it. We could also do that in a later PR if we decide we want to. (And thinking about it again, my 'they are both serialisers' remark was a bit daft, so never mind that. `:P`)

andrevidela:
Actually I'll wait for #141 and update the api to return the errors from the parser instead of `Maybe String`

andrevidela:
@clayrat @epost I'm going to merge this, I've added an issue to track the fact that the JS API need to be updated again after #141 . sounds good?

Issue #138, opened 2019-05-09
elba build missing dependencies


marcosh:
It looks like `elba.toml` is missing some dependencies (e.g. `ziman/idris-bytes`)

andrevidela:
So I've looked into it and I tried to add an `elba.toml` file to the `idris-bytes` projects but it doesn't seem to work out of the box. I've opened an issue on the elba project here https://github.com/elba/elba/issues/54

marcosh:
nice, let's see (if and) what they'll answer

andylokandy:
@andrevidela I think this issue is resolved by https://github.com/elba/elba/commit/4501ea6eb76b1ad2111262b1e5885eb78ffd8dcb and the `include` issue https://github.com/elba/elba/issues/56 is just an improvment and is not necessary here.

andrevidela:
@andylokandy You are right it works now! It think we can close this thanks to you.

andrevidela:
MIght have talked too early, something else is wrong

andylokandy:
What is going wrong?

andrevidela:
@andylokandy nothing to do with elba (afaik) just configuration issues

PR #148, opened 2019-06-22
Add elba support for typedefs-parser

### LINKED ISSUE #147: Make typedefs-parser compile with elba
#138 fixes the build for the _library_ target but we also have a _binary_ target that is currently only compatible with nix builds and manual download of dependencies.

We should add a binary target to the elba manifest (`elba.toml`) and have it compile the `typedefs-parser.ipkg` target.

andrevidela:
Fixes #147 and #138

### DIFF
--- a/elba.toml
+++ b/elba.toml
@@ -4,10 +4,13 @@ version = "0.1.0"
 authors = []
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


andrevidela:
@clayrat Can you check it works on your computer?

clayrat:
@andrevidela yes it does.

PR #149, opened 2019-06-22
WIP - Add tutorial to get started with typedefs

andrevidela:
depends on #148

### DIFF
--- a/TUTORIAL_UNDERSTANDING_TYPEDEFS.md
+++ b/TUTORIAL_UNDERSTANDING_TYPEDEFS.md
@@ -1,0 +1,298 @@
+
+# Tutorial: Understanding Typedefs
+
+This tutorial is aimed at developpers unfamiliar the concepts that drive Typedefs and
+attempts to teach them from the bottom up.
+It is not strictly necessary in order to start using Typedefs but is very useful in
+order to learn how to understand how to achieve certain goals intuitively.
+If you are familiar with functional programming or curious about how Typedefs works, this tutorial
+is for you. Beyond that, it assumes very little knowledge from the reader.
+
+## Algebras
+
+Typedefs is commonly called, in the mathematics jargon, an Algebra. More
+specifically an F-Algebra. Algebras allow us to describe objects with some structure
+using pre-defined _terms_. In our case the terms represent our types and the operation
+represent the shape of the type. Before we dive into types as an algebra
+let us first define and understand what an algebra is in the general sense.
+
+## Terms and operations
+
+You might be familiar with algebra in high school where you are taught to solve equations
+like:
+
+```
+3x + 5 = 17
+```
+
+Where solving means "reduce the equation such that we end up with an expression of the
+shape `x = _`". In high school, `x` meant "here is a number, we don't know anything about
+it but we know there is one" and `=` meant "what is on the left is equal to what is
+on the right".
+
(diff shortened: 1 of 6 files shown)


TUTORIAL_UNDERSTANDING_TYPEDEFS.md
```diff
@@ -0,0 +1,248 @@
...
+```
+
+Which means that an `Optional` type can be one of two values, either `none`, or `some(T)`. The `T` in the second one refers to a type parameter, which mean that a type fits there, we don't know which
+one yet, but any type could work. For example `some(3)` is a valid value with `T = Int`, likewise `some(True)` is also a valid value with `T = Bool`.
+
+**So how many elements does Optional have?**
+
+This one is tricky because there are two answers. Looking at the definition we can reasonably come to the conclusion that there is only two values in the `enum` and therefore only two values for the `Optional` type: `some` and `none`.
+
+Another answer is that it depends on the parameter type. Indeed we already mentionned that `some(3)` and `some(True)` are valid values, if we were to enumerate other values we would have:
```
epost, line 205:
```suggestion
Another answer is that it depends on the parameter type. Indeed we already mentioned that `some(3)` and `some(True)` are valid values. If we were to enumerate other values we would have:
```

TUTORIAL_UNDERSTANDING_TYPEDEFS.md
```diff
@@ -0,0 +1,248 @@
...
+
+- `1` value for `none`
+- since `some(True)` is a value, then `some(False)` also is one
+- since `some(3)` is a value, so is `some(0)`, `some(1)`, and so on are valid values
```
epost, line 209:
```suggestion
- since `some(3)` is a value, so are `some(0)`, `some(1)`, and so on
```

TUTORIAL_UNDERSTANDING_TYPEDEFS.md
```diff
@@ -0,0 +1,248 @@
...
+
+For the last case, if the type parameter is `Int` then we have as many value for the `some` case as there are `Integer` values. So that's `2^32 = 8'589'934'592`, adding the `none` value that's `8'589'934'592 + 1 = 8'589'934'593` values.
+
+As we see, the pattern is that the number of elements of `Optional` is the number of element of the parameter type plus one. So if we use `T` as the type parameter in our `size` syntax
+we have `size(optional<T>) = size(T) + 1`
+
+#### Either/Result values
+
+Finally we are going to look at a type that reprensent two possible alternative when computing
+a value. In Haskell this is the `Either` type, in Scala this type is called `Try`, in Rust it
```
epost, line 223: I think Scala's `Try` isn't quite `Either`. Scala actually has (various) `Either` types, see e.g. https://www.scala-lang.org/api/2.9.3/scala/Either.html from the std lib.

TUTORIAL_UNDERSTANDING_TYPEDEFS.md
```diff
@@ -0,0 +1,248 @@
...
+is `Result`. This type encapuslate one of two possible values that can be of any type. For example the type `Either<String, Int>` has value that can either be `String` or `Int` but we 
+do not know for sure. The construction of such a type is very similar to maybe:
```
epost, line 225:
```suggestion
do not know for sure. The construction of such a type is very similar to `Maybe`:
```

PR #162, opened 2019-07-30
Add name lookup and specialized types syntax

andrevidela:
Implements #165

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,33 +5,33 @@ all: build-all doc-all test-all
 build-all: build-lib install-lib build-rest
 
 build-lib:
-	make build pkg=typedefs
+	make build pkg=typedefs-core
 
 build-rest:
 	make build pkg=typedefs-examples
-	make build pkg=typedefs-parser
-	make build pkg=typedefs-parser-js
+	make build pkg=typedefs-core
+	make build pkg=typedefs-js
 
 doc-all:
-	make doc pkg=typedefs
+	make doc pkg=typedefs-core
 
 test-all:
-	make test pkg=typedefs
+	make test pkg=typedefs-core
 
 install-lib:
-	make install pkg=typedefs
+	make install pkg=typedefs-core
 
 install-all:
 	make install pkg=typedefs
 	make install pkg=typedefs-examples
-	make install pkg=typedefs-parser
-	make install pkg=typedefs-parser-js
+	make install pkg=typedefs-core
+	make install pkg=typedefs-js
 
 clean-all:
 	make clean pkg=typedefs
 	make clean pkg=typedefs-examples
-	make clean pkg=typedefs-parser
-	make clean pkg=typedefs-parser-js
+	make clean pkg=typedefs-core
+	make clean pkg=typedefs-js
 
 build: src
 	idris --build ${pkg}.ipkg
--- a/elba.toml
+++ b/elba.toml
@@ -27,15 +27,15 @@ mods = [ "Typedefs.Names"
        , "Typedefs.Backend.ReasonML"
        , "Typedefs.Backend.Utils"
        ]
-idris_opts = ["-p","contrib"]
+idris_opts = ["-p","contrib","-p","effects"]
 
 [[targets.bin]]
 path = "src"
 name = "typedefs"
 main = "CLI/Main.idr"
-idris_opts = ["-p","contrib"]
+idris_opts = ["-p","contrib","-p","effects"]
 
 [[targets.test]]
 path = "src"
 main = "Typedefs.Test.TypedefsSuite.testSuite"
-idris_opts = ["-p","contrib","--total"]
+idris_opts = ["-p","contrib","-p","effects","--total"]
--- a/src/Typedefs/Backend.idr
+++ b/src/Typedefs/Backend.idr
@@ -12,7 +12,6 @@ import Text.PrettyPrint.WL
 %default total
 %access public export
 
-data CompilerError = RefNotFound String
 
 --TODO added to TParsec
 Traversable NEList where
--- a/src/Typedefs/Backend.idr
+++ b/src/Typedefs/Backend.idr
@@ -12,6 +12,8 @@ import Text.PrettyPrint.WL
 %default total
 %access public export
 
+data CompilerError = RefNotFound String
+
 --TODO added to TParsec
 Traversable NEList where
   traverse f (MkNEList x xs) = [| MkNEList (f x) (traverse f xs) |]
@@ -31,14 +33,14 @@ fromSigma False (S _** _) = Nothing
 ||| @freeVars flag controlling if type definition can have free variables.
 interface ASTGen def type (freeVars : Bool) | def where
   ||| Given a list of `TNamed`, generate their corresponding type signatures.
-  msgType        : ZeroOrUnbounded TNamed freeVars -> type
+  msgType          : ZeroOrUnbounded TNamed freeVars          -> Either CompilerError type
 
   ||| Generate definitions for a list of `TNamed`.
-  generateTyDefs : NEList (ZeroOrUnbounded TNamed freeVars) -> List def
+  generateTyDefs   : NEList (ZeroOrUnbounded TNamed freeVars) -> Either CompilerError (List def)
 
   ||| Generate serialisation and deserialisation term definitions for a
   ||| a `TNamed` and all its helper definitions.
-  generateTermDefs : NEList (ZeroOrUnbounded TNamed freeVars) -> List def
+  generateTermDefs : NEList (ZeroOrUnbounded TNamed freeVars) -> Either CompilerError (List def)
 
 ||| Interface for code generators that can generate code for type definitions and
 ||| type signatures independently of each other, for example Haskell and ReasonML.
@@ -58,17 +60,19 @@ interface CodegenIndep def type | def where
 ||| Use the given backend to generate code for a list of type definitions.
 generateDefs : (def : Type) -> (ASTGen def type fv, CodegenIndep def type) => NEList (n ** TNamed n) -> Maybe Doc
 generateDefs {fv} def tns = 
-  (\nel => 
-    vsep2 $ 
-      (preamble {def}) ::
-      (defSource <$> 
-       (generateTyDefs {def} nel ++ generateTermDefs {def} nel))
-  ) <$> (traverse (fromSigma fv) tns)
+  -- (\nel => 
+  --   vsep2 $ 
+  --     (preamble {def}) ::
+  --     (defSource <$> 
+  --      (generateTyDefs {def} nel ++ generateTermDefs {def} nel))
+  -- ) 
+  ?generateDefsrhs<$> (traverse (fromSigma fv) tns)
 
 ||| Use the given backend to generate code for a list of type signatures.
 generateType : (def : Type) -> (ASTGen def type fv, CodegenIndep def type) => NEList (n ** TNamed n) -> Maybe Doc
 generateType {fv} def tns = 
-  (concatMap (typeSource {def} . msgType {def})) <$> (traverse (fromSigma fv) tns)
+  -- (concatMap (typeSource {def} . msgType {def})) 
+  ?generateTypeRhs <$> (traverse (fromSigma fv) tns)
 
 ||| Interface for code generators that need to generate code for type definitions and
 ||| type signatures at the same time, for example the JSON schema backend.
@@ -81,11 +85,12 @@ interface CodegenInterdep def type where
 ||| Use the given backend to generate code for a list of type definitions.
 generate : (def : Type) -> (ASTGen def type fv, CodegenInterdep def type) => NEList (n ** TNamed n) -> Maybe Doc
 generate {fv} def tns = 
-  (\nel => 
-    sourceCode 
-     ((msgType {def}) <$> nel)
-     (generateTyDefs {def} nel ++ generateTermDefs {def} nel)
-  ) <$> (traverse (fromSigma fv) tns)
+  -- (\nel => 
+  --   sourceCode 
+  --    ((msgType {def}) <$> nel)
+  --    (generateTyDefs {def} nel ++ generateTermDefs {def} nel)
+  -- ) 
+  ?generateRhs <$> (traverse (fromSigma fv) tns)
 
 {-
 record SpecialiseEntry where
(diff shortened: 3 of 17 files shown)


andrevidela:
@clayrat It's finally done, you can start reviewing.

I've left the holes `TermParse.deserialize_bin`, `TermParse.errwhat` unfilled. I will complete them with #166 .

It means that using `(ref bool)` will crash for now. I think it's fine to merge this because it is not a release build, but it should probably be part of at `dev` or `staging` branch instead of master.

(`(ref identifier)` isn't documented anywhere so we can say it unsuported beta software ¯\_(ツ)_/¯ )

@fredrikNordvallForsberg  you can take a look as well but I don't think there is anything special to see here for you, the semantics have not changed.

fredrikNordvallForsberg:
I will take a look in the next few days!

src/Backend/Haskell.idr
```diff
@@ -346,48 +402,68 @@ freshEnvWithTerms : (prefix : String) -> Vect n (HsType, HsTerm)
...
-makeType _    T0          = HsVoid
-makeType _    T1          = HsUnit
-makeType e    (TSum xs)   = foldr1' (\t1,t2 => HsSum t1 t2) (map (assert_total $ makeType e) xs)
-makeType e    (TProd xs)  = HsTuple $ map (assert_total $ makeType e) xs
-makeType e    (TVar v)    = Vect.index v e
-makeType e td@(TMu cases) = HsParam (nameMu cases) $ getUsedVars e td
-makeType e    (TApp f xs) = HsParam (name f) (map (assert_total $ makeType e) xs)
+makeType : Vect n HsType -> TDef n -> LookupM HsType
+makeType _    T0          = pure $ HsVoid
+makeType _    T1          = pure $ HsUnit
```
clayrat:
```suggestion
makeType _    T1          = pure HsUnit
```

src/Backend/Haskell.idr
```diff
@@ -346,48 +402,68 @@ freshEnvWithTerms : (prefix : String) -> Vect n (HsType, HsTerm)
...
-makeType : Vect n HsType -> TDef n -> HsType
-makeType _    T0          = HsVoid
-makeType _    T1          = HsUnit
-makeType e    (TSum xs)   = foldr1' (\t1,t2 => HsSum t1 t2) (map (assert_total $ makeType e) xs)
-makeType e    (TProd xs)  = HsTuple $ map (assert_total $ makeType e) xs
-makeType e    (TVar v)    = Vect.index v e
-makeType e td@(TMu cases) = HsParam (nameMu cases) $ getUsedVars e td
-makeType e    (TApp f xs) = HsParam (name f) (map (assert_total $ makeType e) xs)
+makeType : Vect n HsType -> TDef n -> LookupM HsType
+makeType _    T0          = pure $ HsVoid
```
clayrat:
```suggestion
makeType _    T0          = pure HsVoid
```

src/Backend/JSON.idr
```diff
@@ -27,47 +31,55 @@ nary name = map ((name ++) . show . finToNat) range
...
       , ("additionalProperties", JBoolean False)
-      , ("properties", JObject [(name, assert_total $ makeSubSchema td)])
+      , ("properties", JObject [(name, assert_total !(makeSubSchema td))])
       ]
 
   ||| Generate a schema that requires all of the supplied schemas to match, with each being wrapped in its corresponding name.
-  productSubSchema : Vect k Name -> Vect k (TDef 0) -> JSON
-  productSubSchema names tds = JObject
+  productSubSchema : Vect k Name -> Vect k (TDef 0) -> JSONLookupM JSON
+  productSubSchema names tds = pure $ JObject
```
clayrat:
```suggestion
  productSubSchema names tds = pure JObject
```

src/Backend/JSON.idr
```diff
@@ -27,47 +31,55 @@ nary name = map ((name ++) . show . finToNat) range
...
+    pure $ JObject [("oneOf", JArray . toList $ !(traverseEffect (assert_total makeCase) cases))]
     where
     isMu : TDef n -> Bool
     isMu (TMu _) = True
     isMu _       = False
 
-    makeCase : (Name, TDef 0) -> JSON
-    makeCase (name, td) = JObject
+    makeCase : (Name, TDef 0) -> JSONLookupM JSON
+    makeCase (name, td) = pure $ JObject
```
clayrat:
```suggestion
    makeCase (name, td) = pure JObject
```

src/Backend/JSON.idr
```diff
@@ -117,16 +130,17 @@ makeSchema schema defs = JObject
                   , ("properties", JObject [ ("value", JObject [("oneOf", JArray $ NEList.toList schema)] ) ])
                   ]
 
-generateSchema : TNamed 0 -> JSON
-generateSchema tn = makeSchema (singleton $ makeSubSchema' tn) (evalState (makeDefs' tn) [])
+-- generateSchema : TNamed 0 -> JSON
+-- generateSchema tn = makeSchema (singleton $ makeSubSchema' tn) (runMakeDefM (makeDefs' tn))
```
clayrat: Is this not working?

src/Backend/ReasonML.idr
```diff
@@ -84,64 +95,72 @@ rmlParam : Decl -> RMLType
...
-makeType _     T0         = RMLParam "void" []
-makeType _     T1         = RMLUnit
-makeType e    (TSum xs)   = foldr1' (\t1,t2 => RMLParam "Either" [t1, t2]) (map (assert_total $ makeType e) xs)
-makeType e    (TProd xs)  = RMLTuple . map (assert_total $ makeType e) $ xs
-makeType e    (TVar v)    = either RMLVar rmlParam $ Vect.index v e
-makeType e td@(TMu cases) = RMLParam (nameMu cases) . map (either RMLVar rmlParam) $ getUsedVars e td
-makeType e    (TApp f xs) = RMLParam (name f) (map (assert_total $ makeType e) xs)
+makeType : Env n -> TDef n -> ReasonLookupM RMLType
+makeType _     T0         = pure $ RMLParam "void" []
+makeType _     T1         = pure $ RMLUnit
```
clayrat:
```suggestion
makeType _     T1         = pure RMLUnit
```

src/Backend/JSON.idr
andrevidela: It's never used so I intended to remove it but forgot

src/Typedefs.idr
```diff
@@ -34,13 +35,21 @@ mutual
     ||| Application of a named type to a vector of arguments.
     TApp  : TNamed k -> Vect k (TDef n)           -> TDef n
 
+    ||| A Reference to a named typedef
+    TRef  : Name                                  -> TDef n
```
fredrikNordvallForsberg: For future consideration, I wonder if it would make sense to also have a scoped representation (ie an index of the datatype representing references) of these references in order to only represent and hence having to deal with legal states.

src/Backend/Haskell.idr
```diff
@@ -105,6 +108,60 @@ data Haskell : Type where
...
+||| terms for decoding/encoding the types.
+||| @ n the size of the environment
+TermGen : (n : Nat) -> Type -> Type
+TermGen n t = Eff t [ENV n, SPECIALIZED HsType, ERR]
+
+toTermGen : Either CompilerError a -> TermGen n a
+toTermGen (Left err) = raise err
+toTermGen (Right val) = pure val
+
+Ord k => Default (SortedMap k v) where
```
fredrikNordvallForsberg: Is this only used for the initial states below? If so, maybe clearer to just use `empty` directly there? (Since presumably the initial value being empty matters.)

src/Backend/Utils.idr
```diff
@@ -75,8 +73,53 @@ flattenMus muName = flattenMu [muName]
   flattenMu names (TProd ts)  = assert_total $ TProd $ map (flattenMu names) ts
   flattenMu names (TMu cs)    = assert_total $ TMu $ map (map (flattenMu ((nameMu cs) :: names))) cs
   flattenMu names (TApp f xs) = assert_total $ TApp f (map (flattenMu names) xs)
+  flattenMu names (TRef n)    = TRef n
 
 -- TODO: is this in a library somewhere?
 mapWithIndexA : Applicative m => {n : Nat} -> (Fin n -> a -> m b) -> Vect n a -> m (Vect n b)
```
fredrikNordvallForsberg: Is this now unused? Should it replace `traverseWithIndex` in `Backend.Haskell`?

src/Typedefs.idr
```diff
@@ -149,6 +162,9 @@ mutual
   Ty     tvars (TVar v)    = Vect.index v tvars
   Ty     tvars (TMu m)     = Mu tvars (args m)
   Ty     tvars (TApp f xs) = assert_total $ Ty tvars $ def f `ap` xs -- TODO: could be done properly
+  -- either use a proof that no TRefs exist or a proof that every 
+  -- TRef has a mapping to an Idris type
+  Ty     tvars (TRef n)    = Void
```
fredrikNordvallForsberg: This is where scoped references would be very helpful to do the right thing.

andrevidela: That's a very good point, I don't feel like handling errors at this stage of the pipeline is wise.

We should have a vector that verifies that all those references do actually exist and access them without worrying about throwing errors. The vector should be built in an earlier stage which could throw errors (for example if a specialized type isn't available on the platform we're targetting)

src/Backend/Utils.idr
andrevidela: Good call, I removed it

src/Backend/JSON.idr
andrevidela: I reverted the change because it didn't compile

src/Backend/JSON.idr
andrevidela: I reverted the change because it didn't compile

src/Backend/Haskell.idr
andrevidela: Absolutely, I removed the definition and implemented `runLookupM` in `src/Backend/Utils.idr` and I'm using it instead in the haskell backend.

andrevidela:
@fredrikNordvallForsberg I'll add the indexed lookup when implementing specialised types for Haskell if that's ok with you (#166).

I there anything else before we merge this? @clayrat

src/Backend/Haskell.idr
```diff
@@ -710,32 +805,39 @@ decodeDef {n} t@(TName tname td) =
...
+  generateTyDefs tns = runMakeDefsM $ concat <$> traverseEffect (\(Unbounded tn) => makeDefs' tn) (toVect tns)
+  generateTermDefs tns = runMakeDefsM $ 
+    do gen <- traverseEffect genHaskell (toVect tns)
+       pure $ concat gen
+    where
+      genTerms : TNamed n -> HaskellLookupM (List Haskell)
+      genTerms tn = pure [!(encodeDef tn), !(decodeDef tn)]
+
+      generateNext : TNamed n -> HaskellDefM (List Haskell)
+      generateNext tn = if (name tn) `elem` !('Names :- get)
```
clayrat: Isn't this repeating the definition of `ifNotPresent`?

src/Backend/Haskell.idr
```diff
@@ -346,48 +400,68 @@ freshEnvWithTerms : (prefix : String) -> Vect n (HsType, HsTerm)
...
+                              Nothing => raise $ RefNotFound n
 
 ||| Generate a Haskell type from a `TNamed`.
-makeType' : Vect n HsType -> TNamed n -> HsType
+makeType' : Vect n HsType -> TNamed n -> HaskellLookupM HsType
 makeType' e (TName ""   body) = makeType e body --escape hatch; used e.g. for all non-TApp dependencies of a TApp
-makeType' e (TName name body) = HsParam name $ getUsedVars e body
+makeType' e (TName name body) = pure $ HsParam name $ getUsedVars e body
+
+ifNotPresent' : Name -> HaskellDefM (List Haskell) -> HaskellDefM (List Haskell)
```
clayrat: Is this a specialized version of `Backend.Utils.ifNotPresent` to make life easier for the typechecker, or is there something deeper going on? Either way might be useful to add a comment on why it's here.

src/Backend/Haskell.idr
```diff
@@ -525,31 +595,42 @@ decoderTerm td = encoderDecoderTerm decodeName td
...
-    (nubBy (\ (m ** t), (m' ** t') => heteroEqNamed t t') (go env td))
+  pure $ filter (\ (m ** t) => not $ heteroEq (def t) td)
+    (nubBy (\ (m ** t), (m' ** t') => heteroEqNamed t t') !(go env td))
   where
   mutual
-    goMu : Vect n HsType -> Vect k (String, TDef (S n)) -> List (m ** TNamed m)
-    goMu env tds = concatMap (assert_total $ go ((makeType env (TMu tds))::env) . snd) tds
+    -- Traverse TDef lists recursively with `go` 
+    traverseRec : Vect n HsType -> Vect k (TDef n) -> HaskellLookupM (List (m ** TNamed m))
+    traverseRec env xs = do traversed <-  traverseEffect (assert_total (go env)) xs
```
clayrat:
```suggestion
    traverseRec env xs = concat <$> traverseEffect (assert_total (go env)) xs
```

src/Backend/Haskell.idr
```diff
@@ -525,31 +595,42 @@ decoderTerm td = encoderDecoderTerm decodeName td
...
-    go     env   (TApp {k} t@(TName name td) xs) =
-      let envxs = map (makeType env) xs
-          -- dependencies of the actual td
-          depTd = case td of
-                   TMu tds => goMu envxs tds -- skip the TMu itself
-                   _       => go envxs td
-        in
-      depTd ++ [(k**t)] ++ (concatMap (assert_total $ go env) xs) ++ (concatMap fixup xs)
+    go : Vect n HsType -> TDef n -> HaskellLookupM (List (m ** TNamed m))
+    go     env    T0                             = pure $ []
```
clayrat:
```suggestion
    go     env    T0                             = pure []
```
(same below)

andrevidela:
@clayrat Thanks for your review, I'll fix those, rebase and merge this branch. This will allow me to get started on #166

src/Backend/Haskell.idr
andrevidela: Indeed, fixed

src/Backend/Haskell.idr
andrevidela: It's an artefact of previous attempts, I removed it and replaced it with `Backend.Utils.ifNotPresent`

src/Backend/Haskell.idr
andrevidela: Thanks, fixed

andrevidela:
I had to fiddle a bit with the nix-build after rebasing but it should work now

Issue #177, opened 2019-08-14
Implement decidable equality for closed typdefs.


andrevidela:
In order to use closed typedefs as a category which has decidable equality on its objects we need to implement the `DecEq` interface on `TDef 0`

PR #180, opened 2019-08-20
Typedefs decidable equality

marcosh:
solves #177


clayrat:
In principle we could make the inequality proofs into `Uninhabited` instances, but maybe it's not worth the effort if we're not gonna reuse them anywhere else.

marcosh:
@clayrat I looked at the approach used for `Bool` and `Nat` in `Base` and replicated it. Just tell me if you prefer to switch to `Uninhabited` instances

PR #184, opened 2019-08-20
Fix path for tests

### LINKED ISSUE #182: `elba test` fails
launching `elba test` I get

```
[error] Couldn't build test 0 for typedefs/typedefs 0.1.0 (dir+/home/marcosh/projects/statebox/typedefs)
module Test.TypedefsSuite.testSuite isn't a subpath and doesn't exist under path src
```

andrevidela:
fix #182 

@marcosh Can you tell me if that fixes it for you?


marcosh:
yes, it works now!

Issue #197, opened 2019-12-06
create debug function for `TDef` and `TNamed`


marcosh:
we need a function which return a representation of their complete AST used to define a `TDef` and a `TNamed`, so that we can use that to visually compare definitions

PR #199, opened 2019-12-09
Debug

marcosh:
debug feature to inspect the AST of a TDef and a TNamed.

solves #197

### DIFF
--- a/src/Typedefs/Typedefs.idr
+++ b/src/Typedefs/Typedefs.idr
@@ -355,15 +355,15 @@ mutual
   debugTDefVect : Vect k (TDef n) -> String
   debugTDefVect []        = "[]"
   debugTDefVect (x :: xs) = assert_total $
-    "[" ++ foldr (\elem, acc => acc ++ ", " ++ debugTDef elem) (debugTDef x) xs ++ "]"
+    square $ foldr (\elem, acc => acc ++ ", " ++ debugTDef elem) (debugTDef x) xs
 
   debugMu : Vect k (Name, TDef (S n)) -> String
   debugMu []        = "[]"
   debugMu (x :: xs) = assert_total $
-    "[" ++ foldr (\elem, acc => acc ++ ", " ++ debugNamedMu elem) (debugNamedMu x) xs ++ "]"
+    square $ foldr (\elem, acc => acc ++ ", " ++ debugNamedMu elem) (debugNamedMu x) xs
     where
       debugNamedMu : (Name, TDef (S n)) -> String
-      debugNamedMu (name, tdef) = "(" ++ show name ++ ", " ++ debugTDef tdef ++ ")"
+      debugNamedMu (name, tdef) = parens $ show name ++ ", " ++ debugTDef tdef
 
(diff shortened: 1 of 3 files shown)


src/Typedefs/Typedefs.idr
```diff
@@ -345,3 +345,37 @@ mutual
...
+
+-- we use a named implementation of `Show (Fin n)` to avoid possible conflicts
+[finSimpleShow] Show (Fin k) where
+  show = show . finToNat
+
+mutual
+  debugTDefVect : Vect k (TDef n) -> String
+  debugTDefVect []        = "[]"
+  debugTDefVect (x :: xs) = assert_total $
+    "[" ++ foldr (\elem, acc => acc ++ ", " ++ debugTDef elem) (debugTDef x) xs ++ "]"
```
clayrat: Could reuse the `square` print helper here and below

src/Typedefs/Typedefs.idr
```diff
@@ -345,3 +345,37 @@ mutual
...
+
+  debugMu : Vect k (Name, TDef (S n)) -> String
+  debugMu []        = "[]"
+  debugMu (x :: xs) = assert_total $
+    "[" ++ foldr (\elem, acc => acc ++ ", " ++ debugNamedMu elem) (debugNamedMu x) xs ++ "]"
+    where
+      debugNamedMu : (Name, TDef (S n)) -> String
+      debugNamedMu (name, tdef) = "(" ++ show name ++ ", " ++ debugTDef tdef ++ ")"
```
clayrat: Also `parens` helper here

marcosh:
@clayrat thanks, I integrated your suggestions

src/Typedefs/Typedefs.idr
```diff
@@ -345,3 +345,37 @@ mutual
...
   (TName n t) == (TName n' t')       = n == n' && t == t'
 
+-- Debug ----
+
+-- we use a named implementation of `Show (Fin n)` to avoid possible conflicts
+[finSimpleShow] Show (Fin k) where
+  show = show . finToNat
+
+mutual
+  debugTDefVect : Vect k (TDef n) -> String
```
andrevidela: Why can't this be implemented with `show` on vectors?

clayrat: Probably because of `debugTDef` reference

clayrat: Probably because of `debugTDef`

marcosh: ah, this could actually be a named implementation of the `Show` interface, let me try if it works

marcosh: actually this does not seem to work easily because `show :: Vect n a -> String` expects an implementation of `Show (Vect n a)` and not an implementation of `Show a`

clayrat: yeah, you'll need to make everything named instances and thread them through

marcosh: ok, so I think this is ready to be merged now

PR #203, opened 2019-12-21
Fix ref parsing

### LINKED ISSUE #200: Fix Termparse for specialised types
The current parser doesn't has been commented out since it can't handle reference indexes after #191 .

The solution would be to state the list of used references and add them to the context for parsing:

```
(specialised int)
(specialised List)
(name listInt ((ref 1) (* (ref 0) (var 2))))
```

should be 

```
type ListInt x0 = List (Int , x0)
```

andrevidela:
fix #200

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -32,7 +32,7 @@
 #    --arg pkgs 0
 #
 # c) use nixpkgs from an URL
-#    --arg pkgs 0 -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/18.09.tar.gz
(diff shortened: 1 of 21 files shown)


clayrat:
TParsec@nixpkgs is now updated, but we're still blocked by nixpkgs unstable builds failing, adding a link to track this: https://hydra.nixos.org/job/nixpkgs/trunk/unstable/all

clayrat:
There's now a functioning nixpkg unstable build with our update, but https://nixos.org/channels/nixpkgs-unstable isn't still updated to that one, I guess we now need to wait for this.

andrevidela:
FINALLLYYYY

PR #204, opened 2019-12-25
Index free syntax

andrevidela:
This PR add support for a new index-free syntax which more closely resembles ml-style languages with data constructors:

```
List a := nil : 1 | cons : a * List a
```

### DIFF
--- a/src/Typedefs/Syntax/Compiler.idr
+++ b/src/Typedefs/Syntax/Compiler.idr
@@ -9,6 +9,8 @@ import Effect.State
 import Typedefs.Typedefs
 import Typedefs.Syntax.AST
 import Typedefs.Names
+import Typedefs.Backend.Utils
+import Typedefs.Parse
 
 %default total
 
@@ -51,7 +53,7 @@ fromNat (S (S n)) = TSum (replicate (S (S n)) T1)
 mutual
   compilePower : AST.Power -> CompilerM (Either Nat (TDef Z))
   compilePower (PLit n) = pure (Left n)
-  compilePower (PEmb x) = pure (Right !(compileFactor x))
+  compilePower (PEmb x) = pure (Right !(compileExpr x))
 
   compileFactor : AST.Factor -> CompilerM (TDef Z)
   compileFactor (FEmb x) = do Left n <- compilePower x
@@ -63,18 +65,18 @@ mutual
          Left p => pure $ (Typedefs.pow p f)
          Right def => pure def
 
-compileTerm : AST.Term -> CompilerM (TDef Z)
-compileTerm (TEmb x) = compileFactor x
-compileTerm (TMul x y) = do t <- compileTerm x
-                            f <- compileFactor y
-                            pure $ (TProd [t, f])
+  compileTerm : AST.Term -> CompilerM (TDef Z)
+  compileTerm (TEmb x) = compileFactor x
+  compileTerm (TMul x y) = do t <- compileTerm x
+                              f <- compileFactor y
+                              pure $ (TProd [t, f])
 
-compileExpr : AST.Expr -> CompilerM (TDef Z)
-compileExpr (EEmb x) = compileTerm x
-compileExpr (ESum x y) = do (e) <- compileExpr x
-                            (t) <- compileTerm y
-                            pure $ TSum [e, t]
-compileExpr (Ref x) = pure $ (TRef x)
+  compileExpr : AST.Expr -> CompilerM (TDef Z)
+  compileExpr (EEmb x) = compileTerm x
+  compileExpr (ESum x y) = do (e) <- compileExpr x
+                              (t) <- compileTerm y
+                              pure $ TSum [e, t]
+  compileExpr (Ref x) = pure $ (TRef x)
 
 plusSuccSucc : (li, ri : Nat) -> plus (S li) (S ri) = S (S (plus li ri))
 plusSuccSucc li ri = cong {f=S} $ sym $ plusSuccRightSucc li ri
@@ -167,21 +169,34 @@ mutual
                                       pure (S l ** (prf, TVar index, newCtx))
   indexify names ctx _ = raise "shouldn't happen, lol typechecking amirite"
 
+||| Given the name of an enum and the name of its argument construct a correctly indexed TDef
 indexEnum : (name : String) -> (args : List String)
          -> TDef Z -> Eff (n ** TDef n) [EXCEPTION String]
 indexEnum name args tdef = do (n ** (_, newTDef, ctx)) <- indexify args ["b"] tdef
                               pure (n ** newTDef)
+
+checkDefs : Vect n a -> CompilerM (k ** Vect (2 + k) a)
+checkDefs [x, y] = pure (Z ** [x, y])
+checkDefs (x :: xs) = do (n ** ns) <- checkDefs xs
+                         pure (S n ** x :: ns)
+checkDefs _ = raise "enum doesn't have at least two constructors"
+
+maximize : Vect (S k) (n ** TDef n) -> (m ** Vect (S k) (TDef m))
+maximize vect = let (n ** max) = toVMax vect in
+                    (n ** map (\(_ ** (lte, td)) => weakenTDef td n lte) (fromVMax max))
 
 ||| Compile an enum syntax down to a TMu
 ||| @name the name given to the type
 ||| @args the names used as type parameters
 ||| @constructors the list of constructors for each value
 compileEnum : (name : String) -> (args : List String)
-           -> (constructors : List (String, TDef Z))
+           -> (constructors : Vect m (String, TDef Z))
            -> CompilerM (n ** TNamed n)
 compileEnum name args constructors = do
-  let asVect = fromList $ map (map $ indexEnum name args) constructors
-  pure $ ?what
+  (k ** checkedDefs) <- checkDefs (constructors)
+  defs <- traverseEffect (\(n, def) => indexEnum name args def ) checkedDefs
+  let (i ** maximized) = maximize $ defs
+  pure $ (i ** TName name $ TSum {k} maximized)
 
 compileDef : AST.TopLevelDef -> CompilerM (n ** TNamed n)
 compileDef (MkTopLevelDef (MkDefName n (x :: xs)) (Simple y)) =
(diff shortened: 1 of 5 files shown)


Typedefs_Syntax.md
```diff
@@ -0,0 +1,138 @@
+
+# Typedefs Syntax
+
+This document explains the brand new Typedefs Syntax for Typedefs.
+
+## Building blocs
```
epost, line 6:
```suggestion
## Building blocks
```

Typedefs_Syntax.md
```diff
@@ -0,0 +1,138 @@
...
+
+This new Typedefs syntax only has a couple Building blocks: `0`, `1`, `+`, `*` and `^`
```
epost, line 8:
```suggestion
This new Typedefs syntax only has a couple of building blocks: `0`, `1`, `+`, `*` and `^`.
```

Typedefs_Syntax.md
```diff
@@ -0,0 +1,138 @@
...
+
+- `0` represents the empty type, an irrepresentable type with no inhabitant. Most useful when used
+with `+` and type parameters
```
epost, line 11:
```suggestion
- `0` represents the empty type, an irrepresentable type with no inhabitants. Most useful when used
with `+` and type parameters.
```

Typedefs_Syntax.md
```diff
@@ -0,0 +1,138 @@
...
+- `1` represents the unit type. It has a single inhabitant which is trivially constructible.
+- `+` Allows you to pick a single value between two possible types.
+- `*` Allows you to package two types together in a single value.
+- `^` Multiple uses of `*`:  `a * a * a = a ^ 3`
```
epost, line 15: Suggestion: mention operator names? (Coproduct / disj. union, product / pair / tuple, power, etc.)

Typedefs_Syntax.md
```diff
@@ -0,0 +1,138 @@
...
+
+For more details look at the [Theoretical tutorial](TUTORIAL_UNDERSTANDING_TYPEDEFS.md)
+
+You can build Typedefs with those simple building blocs by binding them to a name with `:=` and
```
epost, line 19:
```suggestion
You can build Typedefs with those simple building blocks by binding them to a name with `:=` and
```

Typedefs_Syntax.md
```diff
@@ -0,0 +1,138 @@
...
+Either a b := a + b;
+```
+
+This example defines a Typedef `Maybe` which takes a type parameter `a` and is used as the second
+alterntive to `Maybe`. `Either` takes two type parameters and says that a value is either the first
+one or the second one.
+
+## Type application
+
+If you want to refer to a type that's already been defined you can simply refer to it's name like
```
epost, line 46:
```suggestion
If you want to refer to a type that's already been defined you can simply refer to its name like
```

Typedefs_Syntax.md
```diff
@@ -0,0 +1,138 @@
...
+Indeed we require self-referential types to name each of their constructors using the following
+syntax:
+
+```
+List a := Nil : 1 | Cons : (a * List a)
+```
+
+Where the `|` operator acts like `+` when dealing with named constructors. This syntax is reminicent
+of ML-like languages that allow you to define datatypes with multiple constructors, this is also
+the case here:
```
epost, line 79:
```suggestion
Where the `|` operator acts like `+` when dealing with named constructors. This syntax is reminiscent
of ML-like languages, which allow you to define data types with multiple constructors. This is also
the case here:
```

Typedefs_Syntax.md
```diff
@@ -0,0 +1,138 @@
...
+
+```
+Tree a b := Tensor   : (Tree a b) * (Tree a b)
+          | Sequence : (Tree a b) * (Tree a b)
+          | Sym      : a * a
+          | Id       : a
+          | Mor      : b
+```
+
+(An example directly taken from our `FSM-oracle` project)
```
epost, line 89:
```suggestion
(This example was directly taken from our `FSM-oracle` project.)
```
Suggestion: link to the relevant FSM-oracle code?

Typedefs_Syntax.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Named fields
+
+You can also use the `|` syntax to include names for your type constructors if you find that `+`
+obscures the meaning of your constructors. However, `|`-declarations cannot be nester, wheras
```
epost, line 94:
```suggestion
obscures the meaning of your constructors. However, `|`-declarations cannot be nested, whereas
```

Typedefs_Syntax.md
```diff
@@ -0,0 +1,138 @@
...
+
+Maybe a = Either 1
+```
+
+
+## End remarks
+
+If you are familiar with typedefs, you will notice that `mu`, `app` and `var` are
+missing, this is because the new syntax mostly superseed those constructors and compiles
+down to a representation that uses them.
```
epost, line 134:
```suggestion
If you are familiar with Typedefs, you will notice that `mu`, `app` and `var` are
missing. This is because the new syntax mostly supersedes those constructors and compiles
down to a representation that uses them.
```

src/CLI/CommandLineParser.idr
```diff
@@ -4,6 +4,23 @@ import Options.Applicative
...
+public export
+data InputFormat =
+
+  ||| Lisp syntax is the traditional lisp-like syntax for Typedefs
+  Lispy |
+
+  ||| IndexFree syntax is the new Typedefs syntax
+  ||| That looks more like ml-style data declaration
+  ||| It has some limitations compared to lispy, but is much more
+  ||| familiar to use.
```
epost, line 16:
```suggestion
  ||| IndexFree syntax is the new Typedefs syntax that looks more like ML-style data declarations.
  ||| It has some limitations compared to the lispy one, but is much more familiar to use.
```

epost: Suggestion: shouldn't we call it `MLStyle` instead then?

src/Typedefs/Syntax/Compiler.idr
```diff
@@ -0,0 +1,234 @@
...
+    isVarList' (TVar var :: xs) idx = if toNat var == idx then isVarList' xs (S idx)
+                                                          else False
+    isVarList' [] _ = True
+    isVarList' _ _ = False
+
+isRefName : String -> TNamed n -> Bool
+isRefName n (TName name (FRef rn)) = n == name && rn == n
+isRefName _ _ = False
+
+-- The next three functions could probably be asbtracted somehow
```
epost, line 146:
```suggestion
-- The next three functions could probably be abstracted somehow
```

src/Typedefs/Syntax/Compiler.idr
```diff
@@ -0,0 +1,234 @@
...
+ComputeEffect : {a : Nat -> Type} -> (n ** a n) -> List EFFECT
+ComputeEffect (l ** _) = [EXCEPTION String, STATE (Vect l String)]
+
+-- convert a Nat into a sum of T1s
+fromNat : Nat -> TDef n
+fromNat Z = T0
+fromNat (S Z) = T1
+fromNat (S (S n)) = TSum (replicate (S (S n)) T1)
+
+-- Given a proof than an element is in the list, make a TVar with its index in the list
```
epost, line 32:
```suggestion
-- Given a proof that an element is in the list, make a TVar with its index in the list
```

src/Typedefs/Syntax/Compiler.idr
```diff
@@ -0,0 +1,234 @@
...
+makeElem : {e : String} -> Data.List.Elem e ls -> TDef (length ls)
+makeElem Here = TVar FZ
+makeElem (There later) = shiftVars $ makeElem later
+
+-- If the string is an element in the list, make a TVar out of it otherwise return a ref
```
epost, line 37:
```suggestion
-- If the string is an element in the list, make a TVar out of it; otherwise return a ref
```

src/Typedefs/Syntax/Compiler.idr
```diff
@@ -0,0 +1,234 @@
...
+                                    t <- compileTerm names y
+                                    pure $ TSum [e, t]
+
+  compileAppli : (names : List String) -> AST.Appli -> CompilerM (TDef (length names))
+  compileAppli names (AEmb val) = compileVal' names val
+  compileAppli names (App c a) = do (l ** (contructor, args)) <- compileArgs [] c a
+                                    pure $ TApp (TName (show contructor) contructor) args
+    where
+      ||| Returns left if the application just wraps an expression
+      ||| otherwise return the details of the app
```
epost, line 81:
```suggestion
      ||| otherwise returns the details of the app
```

src/Typedefs/Syntax/Compiler.idr
```diff
@@ -0,0 +1,234 @@
...
+      compileArgs : (acc : List (TDef (length names))) -> AST.Appli -> Val ->
+                    CompilerM ((l ** (TDef l, Vect l (TDef (length names)))))
+      compileArgs acc (AEmb val) arg = do
+        arg' <- compileVal' names arg
+        let acc' = arg' :: acc
+        -- we need to reduce here because the type we apply might use less indices than those
```
epost, line 87: s/less/fewer/?

Typedefs_Syntax.md
andrevidela: It would be great but the code hasn't been merged yet.

src/CLI/CommandLineParser.idr
andrevidela: I guess? I feel like the ML-style is the marketing term while the original intent for it was to avoid having to manually track debrujin indices

src/Typedefs/Test/IndexFreeTests.idr
```diff
@@ -0,0 +1,99 @@
...
+                                            (EEmb $ TEmb $ FEmb $ AEmb $ PRef "a")
+                                            (TEmb $ FEmb $ AEmb $ PRef "b"))
+                                          (FEmb $ App (App (AEmb $ PRef "BinaryTree")
+                                                           (PRef "a"))
+                                                      (PRef "b")))
+                   ])) [])
+
+
+--    "List a := Nil : 1 | Cons : a + List a"
+--      `tdefProgramShouldParseAs`
```
andrevidela, line 91: should remove those tests

src/Typedefs/Test/IndexFreeTests.idr
```diff
@@ -0,0 +1,99 @@
...
+    it "BinaryTree a b := Left : a | Right : b | Node : (a + b) * BinaryTree a b;" $
+       (parseDefList "BinaryTree a b := Left : a | Right : b | Node : (a + b) * BinaryTree a b;")
+         `shouldBe`
+           Just (MkNEList (MkTopLevelDef (MkDefName "BinaryTree" ["a", "b"])
+             (Enum [ ("Left", EEmb $ TEmb $ FEmb $ AEmb $ PRef "a")
+                   , ("Right", EEmb $ TEmb $ FEmb $ AEmb $ PRef "b")
+                   , ("Node", EEmb $ TMul (TEmb $ FEmb $ AEmb $ PEmb $ ESum
+                                            (EEmb $ TEmb $ FEmb $ AEmb $ PRef "a")
+                                            (TEmb $ FEmb $ AEmb $ PRef "b"))
+                                          (FEmb $ App (App (AEmb $ PRef "BinaryTree")
```
andrevidela, line 84: Should align

Typedefs_Syntax.md
andrevidela: fixed

src/Typedefs/Syntax/Compiler.idr
andrevidela: fixed

src/CLI/CommandLineParser.idr
andrevidela: I've renamed it ML style since it was either that, or rename the other "SExpr"

andrevidela:
The merge for this is going to take a bit longer than expected

andrevidela:
Actually this is going to take a lot more work than expected. Indeed this is not compatible anymore with the current master in which we only allow to generate code for TDefs with _resolved_ references.

This [commit](https://github.com/typedefs/typedefs/commit/2cba55eb17cc4b69d453c7cf64a0315665f6f6ed), is the culprit and makes the change from `TNamed` to `TNamedR`. I tried using the existing references compiler but the type signatures don't quite match.

So either we rewrite the reference compiler or we rewrite the syntax compiler. Eitherway, it's not a trivial and easy task

Issue #206, opened 2020-02-04
Purescript


andrevidela:
Add support for a purescript backend with both term definition and type definitions.

Could probably be a straight port from the Haskell backend

clayrat:
Should try porting Haskell code using https://github.com/rightfold/purescript-bytestrings

Issue #143, opened 2019-06-08
Split the API between generateType and generateDef


andrevidela:
Currently `generateDef` sometimes return the term serializers and deserializers (for Haskell for example) or the data type definition (in other backends). There should probably be two functions for those two functionalities.

This could also allow to add new options to generate types and serializers, typically we could generate all datatype signatures in a file and serializers in another. Or only generate serializers without datatypes because they use existing types.

This would be a good opportunity to split the haskell backend into two parts, one for types and one for terms.

clayrat:
We already sort of have it, with `ASTGen` having two separate methods: `generateTyDefs` and `generateTermDefs`, it's just that the latter returns empty output for everything except Haskell. We only glue everything back together in [generateDefs](https://github.com/typedefs/typedefs/blob/master/src/Typedefs/Backend.idr#L65). I think the main part missing for this is the ability to generate multiple files and imports within them?

clayrat:
@andrevidela Also which branch should I ideally use a base, I'm guessing [fix-ref-parsing](https://github.com/typedefs/typedefs/tree/fix-ref-parsing)?

andrevidela:
yes [fix-ref-parsing](https://github.com/typedefs/typedefs/tree/fix-ref-parsing),

and yes we already "sort of have it" but it's really not intuitive. If you look at the JS main file you can see how degenerate the API is. We use `generate`, `generateDefs` but `generateDefs` doesn't generate termdefs for ReasonML, only for haskell.

Ideally we would use `generateDefs` for haskell and `generateTyDefs` for the two others.

Also we need to split up the _Type_ generation for Haskell and the _term_ generation. They're both in the same file and it's a huge mess.

andrevidela:
So after the cleaning up here is what we probably need

```idris
interface GenTypes typeAST where
    genTypes : TopLevelDef -> Either CompilerError typeAST

interface GenTerms termAST where
    genTerms : TopLevelDef -> Either CompilerError termAST

interface GenUtils termAST where
    genUtils : TopLevelDef -> termAST

interface RenderSource languageAST where
    render : languageAST -> Doc

renderFile : (GenTypes ast, GenTerms ast, GenUtils ast, RenderSource ast) 
          => TopLevelDef -> Either CompilerError Doc
```

I guess we can salvage `CodegenIndep` to replace `GenUtils` and `renderFile`

PR #207, opened 2020-02-28
Split api

andrevidela:
Laying the groundwork for #143 and #206 

There shouldn't be anything too controversial here except choices about how to split up files and uses of re-exports


clayrat:
Looks cool overall, although the distinction between `Data`, `Type` and `Term` is a bit confusing, should we maybe get rid of `Haskell.Data` and move the definitions into the other two?

andrevidela:
What do you think or renaming `Type` and `Term` to `Typegen` and `Termgen` ?

What do you think about re-exporting `Haskell.Data` in `Haskell.Termgen` instead?

andrevidela:
Wait, do we need to expose `Typegen`, `Termgen` and `Data` in the library? I assumed just having the generating functions would be enough

andrevidela:
@clayrat ?

PR #216, opened 2020-03-28
Fix indexing issues for inline application

### LINKED ISSUE #211: Serialization of terms constructed out of nested Mu's crashes
E.g., `serialize [] [] (TList `ap` [TNat]) (fromList {tdef=TNat} $ map fromNat [3])`

clayrat:
Fixes #211

### DIFF
--- a/src/Typedefs/Test/TermParseWriteTests.idr
+++ b/src/Typedefs/Test/TermParseWriteTests.idr
@@ -1,6 +1,7 @@
 module Typedefs.Test.TermParseWriteTests
 
 import Typedefs.Typedefs
+import Typedefs.Library
 import Typedefs.Names
 import Typedefs.TermParse
 import Typedefs.TermWrite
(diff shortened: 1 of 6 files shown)


src/Typedefs/Typedefs.idr
```diff
@@ -110,27 +110,41 @@ square s = "[" ++ s ++ "]"
...
+ext : Ren n m -> Ren (S n) (S m)
+ext s  FZ    = FZ
+ext s (FS x) = FS (s x)
+
+rename : Ren n m -> TDef' n b ->  TDef' m b
+rename r          T0                     = T0
+rename r          T1                     = T1
+rename r         (TSum ts)               = assert_total $ TSum $ map (rename r) ts
+rename r         (TProd ts)              = assert_total $ TProd $ map (rename r) ts
+rename r         (TMu cs)                = assert_total $ TMu $ map (map $ rename (ext r)) cs
```
andrevidela: So this is where the magic happens

src/Typedefs/Test/TermParseWriteTests.idr
```diff
@@ -19,6 +19,27 @@ roundtrip1 td x = deserializeBinaryClosed td $ serializeBinaryClosed td x
...
+fromNat  Z    = Inn (Left ())
+fromNat (S n) = Inn (Right $ fromNat n)
+
+toList : Ty [] (TList `ap` [tdef]) -> List (Ty [] tdef)
+toList (Inn (Left ()))        = Nil
+toList (Inn (Right (hd, tl))) = ignoreShift hd :: toList tl
+
+fromList : List (Ty [] tdef) -> Ty [] (TList `ap` [tdef])
+fromList  Nil      = Inn (Left ())
+fromList (x :: xs) = Inn (Right (addShift x, fromList xs))
```
andrevidela: Shouldn't this be somewhere else? It would be nice if we could expose it in as a "typedefs standard library"

andrevidela (changes requested):

src/Typedefs/Test/TermParseWriteTests.idr
clayrat: Ok, where do you think we should put this? We already have `examples` which cover both of these, but they're in a different subproject.

andrevidela: How about : `src/Typedefs/Library`? 
we could have `Library.idr` be the standard library and subfolders be more specific imports like
`src/Typedefs/Library/JSON.idr` be all the functions related to using the JSON serialisation format

clayrat: Ok, should I move most of `examples` there too?

clayrat: I'm thinking about `bit`, `byte`, `maybe2` + helpers, `either2` + helpers, `list` and `nat`

andrevidela: mhh probably just Maybe and the generic stuff, `bit` and `byte` doesn't really have a practical purpose do they? Maybe have `Bool` instead of `bit`

andrevidela:
Looks good!

andrevidela (approved):

PR #219, opened 2020-04-20
Add AlphaTy, BetaTy and GammaTy

### LINKED ISSUE #215: Fix ty with idris backend when using references and parameterised types
we can't to dthis

```
TList : TDef 1
TList = RRef 0

instanceLists : (Ty [List] TList) Nat
instanceList = [1,2,3]
```
because `Ty` expects a vector of `Type` and `List` is of type `Type -> Type`

andrevidela:
fix #215 

Those functions allow to either partially apply a TDef with its
Idris Type representation or use a type constructor to fill one of
its holes.

### DIFF
--- a/src/Typedefs/Typedefs.idr
+++ b/src/Typedefs/Typedefs.idr
@@ -218,6 +218,46 @@ mutual
     assert_total $ Ty tvars $ def `ap` xs -- TODO: could be done properly
   Ty     tvars (RRef i)                = Vect.index i tvars
 
+TypeConstructor : Nat -> Type
+TypeConstructor Z = Type
+TypeConstructor (S n) = Type -> TypeConstructor n
+
+ApplyVect : Vect n Type -> TypeConstructor n -> Type
+ApplyVect [] c = c
+ApplyVect (x :: xs) c {n = S k} = ApplyVect xs (c x)
+
+ReverseVect : (mkType : Vect n Type -> Type) -> TypeConstructor n
+ReverseVect mkType {n = Z} = mkType []
+ReverseVect mkType {n = (S k)} = \arg => ReverseVect (popHead mkType arg)
+  where
+    popHead : (Vect (S n) Type -> Type) -> Type -> Vect n Type -> Type
+    popHead f t ts = f (t :: ts)
+
+||| Converts a typedefs of `n` free variables into a type constructor that expects n arguments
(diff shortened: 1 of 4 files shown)


src/Typedefs/Typedefs.idr
```diff
@@ -218,6 +218,45 @@ mutual
...
+ApplyVect (x :: xs) c {n = S k} = ApplyVect xs (c x)
+
+ReverseVect : (mkType : Vect n Type -> Type) -> TypeConstructor n
+ReverseVect mkType {n = Z} = mkType []
+ReverseVect mkType {n = (S k)} = \arg => ReverseVect (popHead mkType arg)
+  where
+    popHead : (Vect (S n) Type -> Type) -> Type -> Vect n Type -> Type
+    popHead f t ts = f (t :: ts)
+
+AlphaTy : TDefR n -> TypeConstructor n
```
clayrat, line 236: Should we add a docstring here as well?

src/Typedefs/Typedefs.idr
```diff
@@ -218,6 +218,45 @@ mutual
...
+||| Given a vector of type constructors and a TDef construct the Idris type
+||| from it using the second vector to instanciate the constructors
+BetaTy : (types : Vect n (k ** TypeConstructor k)) -> TDefR n
+     -> Vect (NatSum types) Type -> Type
+BetaTy types tdef args = Ty (constructTypes types args) tdef
+
+||| Given a list of type constructors fill the holes with them and return a new type constructor
+GammaTy : (types : Vect n (k ** TypeConstructor k)) -> TDefR n
+     -> TypeConstructor (NatSum types)
+GammaTy types tdef {n} = ReverseVect {n=NatSum types} (BetaTy types tdef)
```
clayrat, line 258: Doesn't look like `{n}` is needed here.

andrevidela:
@clayrat I feel like we're getting very close to that since I was looking at how to implement application for type constructor, worth mentionning during the meeting.

src/Typedefs/Typedefs.idr
andrevidela: Fixed

src/Typedefs/Typedefs.idr
andrevidela: Added, thanks!

PR #222, opened 2020-04-23
WIP - add strengthen

### LINKED ISSUE #221: Add ignoreWeaken (reverse of ignoreShift)
We need a way to also drop unused variable that are prepended to context, so something like the following:

`Ty (l++r) (weakenTDef tdef (length r) prf) -> Ty l tdef`

andrevidela:
fix #221 
I added strenghten as a…postulate! Good enough? @clayrat


andrevidela:
I tried impementing it, and as usual Mu is being a pest

andrevidela:
yeah this is garbage

PR #225, opened 2020-05-19
Change Ty to take a specialisation context in argument

### LINKED ISSUE #220: Fix Type Application for references in Idris Type Interpretation
This doesnt work

```idris
test : GammaTy [(1** List)] ((RRef 0) `ap` TProd [TVar 0, TVar 0]) Nat
test = [(1,2)]
```

Indeed `(RRef 0) `ap` TProd [TVar 0, TVar 0]` is executed to be `TProd [TVar0, TVar0]` because `ap` simply replaces `RRef 0` by its argument.

-------------

To expand on that more here is the use case:

We want to use _specialised_ versions of `Nat` and `List` so that we do not have to deal with their `Mu` notation in their term representation. For this, we simply represent `Nat` as being `TNat` which are references to the index 0, likewise for `List`.

```
TNat : TDef 2
TNat = Ref 1

TList : TDef 2
TList = Ref 0
```

We now want to have a pair of lists of nats for this we use the existing constructor `TProd` to construct our pair

```
ListOfPairs : TDef 2
ListOfPairs = TList `ap` TProd [TNat,  TNat]
```

However this definition won't work because `TList` is of type `TDefR 2` and therefore requires 2 arguments to be applied, we're only giving it one. But moreover, this use of `ap` is completely wrong, since it will simply replace the reference `RRef 0` by `TNat` which is not what we expect. 

This is due to how variables and references behave when they are used as type constructors and are applied to something else. You can't apply them to anything because they will themselves be substituted by the values given in argument, but that is not what we mean. What we mean is that they need to say as variables, until they are instanciated as a type constructor in their target language

```
PairOfLists : Ty [List, Nat] (TList `ap` TProd [TNat, TNat])
PairOfLists = [(1,4), (2,5), (3,6)]
```

---------------

This use case appears in fsm-oracle where executions are defined like so: 
```
TNat : TDefR 3
TNat = RRef 0

TListPair : TDefR 3
TListPair = RRef 1

TListNat : TDefR 3
TListNat = RRef 2

TPetriExec = TProd [TProd [TNat , TListPair], TListNat,  TTree]

FromExec : Ty [Nat, List (List Nat, List Nat), List Nat] TPetriExec -> …
```

But ideally should be defined as

```
TPetriExec = TProd [TProd [TNat , TList `ap` TProd [TNat, TNat]], TList `ap` TNat,  TTree]
FromExec : Ty [Nat, List] TPetriExec -> …
```

andrevidela:
Attempt to fix #220 without too much fuss

## How it works

The new `Ty'` function takes a `SpecialList` as its first new argument. This list is a mapping from `TDef n` to `TypeConstructor n`. This replacement is only done on `Mu` and `TApp` since those are the only constructors that have argument vectors. The vectors are then applied to the `TypeConstructor n` we found. If no type constructor is found, the type is applied without replacement as before.

#### Reference to list

```idris
SContext : SpecialList
SContext = [(1 ** (TList, List))]

referToNativeList : Ty' SContext [Nat] TList
referToNativeList = [1,2,3]
```

#### Reference to non-constructors

```idris
TNat : TDefR 0
TNat = TMu [("Nat", T0)] -- uninhabited mu

SContext : SpecialList
SContext = [(0 ** (TNat, Nat))]

referToNat : Ty' SContext [] TNat
referToNat = S Z
```

#### Nested references

```idris
TNat : TDefR 0
TNat = TMu [("Nat", T0)] -- uninhabited mu

SContext : SpecialList
SContext = [ (0 ** (TNat, Nat))
           , (1 ** (TList, List))
           ]
referToListPair : Ty' SContext []  (TList `ap` TProd [TList `ap` TNat, TList `ap` TNat])
referToListPair = [([], [1]), ([1,2,3,4], [4,3,2,1])]
```

## Limitations

I haven't implemented anything with Termwrite or TermParse so I don't know how those will be affected. Currently, since the change is backward compatible, those assume the replacement context is empty.


andrevidela:
@clayrat @fredrikNordvallForsberg Could you two tell me what you think of this? I quickly put this together yesterday and it seems to work fine. Would you say it's a satisfactory solution to allowing references with kinds `* -> *` ?

src/Typedefs/Idris.idr
```diff
@@ -0,0 +1,260 @@
...
+  where
+    popHead : (Vect (S n) Type -> Type) -> Type -> Vect n Type -> Type
+    popHead f t ts = f (t :: ts)
+
+
+mutual
+  data Mu' : SpecialList -> Vect n Type -> TDef' (S n) a -> Type where
+    Inn : Ty' sp (Mu' sp tvars m :: tvars) m -> Mu' sp tvars m
+
+  Tnary' : SpecialList -> Vect n Type -> Vect (2 + k) (TDefR n) -> (Type -> Type -> Type) -> Type
```
fredrikNordvallForsberg, line 50: would it make sense to pull the `SpecialList` out into a
```idris
parameters (rep : SpecialList)
```
since I don't think it changes in any of the calls? (It's tempting to do the same for `Mu'`, but then I think it can't be implicit in the type of the constructor?)

src/Typedefs/Idris.idr
```diff
@@ -0,0 +1,260 @@
...
+    Ty' rep tvars (TMu m) | Nothing = Mu' rep tvars (args m)
+    Ty' rep tvars (TMu m) | Just constr = ApplyVect tvars constr
+
+  -- For application we first make a lookup in our mapping from TDef to Type constructors
+  Ty' rep tvars (TApp (TName _ def) xs) with (custLookup rep def)
+    -- If we can't find anything, simply apply the type normally
+    Ty' rep tvars (TApp (TName _ def) xs) | Nothing = assert_total $ Ty' rep tvars $ def `ap` xs
+    -- If we find a match, check the length of the arguments match the arity of the type constructor
+    -- we found.
+    Ty' rep tvars (TApp (TName _ def) xs) | Just (constr) =
```
fredrikNordvallForsberg, line 78:
```suggestion
    Ty' rep tvars (TApp (TName _ def) xs) | Just constr =
```

src/Typedefs/Idris.idr
```diff
@@ -0,0 +1,260 @@
...
+        let args = map (assert_total $ Ty' rep tvars ) xs in ApplyVect args (constr)
```
fredrikNordvallForsberg, line 79:
```suggestion
        let args = map (assert_total $ Ty' rep tvars) xs in ApplyVect args constr
```

src/Typedefs/Idris.idr
```diff
@@ -0,0 +1,260 @@
...
+  custLookup ((n ** (key, val)) :: xs) item {w} | No _ = custLookup xs item
+  custLookup ((n ** (key, val)) :: xs) item {w=n} | Yes Refl =
+    if item == key then Just val -- index & key match, return value and its index
+                   else custLookup xs item -- index match but values differ
+
+TypeConstructor : Nat -> Type
+TypeConstructor Z = Type
+TypeConstructor (S n) = Type -> TypeConstructor n
+
+ApplyVect : Vect n Type -> TypeConstructor n -> Type
```
fredrikNordvallForsberg, line 25: I think it makes sense to swap the order of the arguments, cf `f(x)` vs `(x)f`

src/Typedefs/Idris.idr
```diff
@@ -0,0 +1,260 @@
...
+
+-------------------------------------------------------------------------------
+-- Type Equality                                                             --
+-------------------------------------------------------------------------------
+-- Note: Equality on Types only work when the specialisation context is empty--
+-------------------------------------------------------------------------------
+
+mutual
+
+  eqMu : (tvars : Vect n (a : Type ** a -> a -> Bool)) -> (td : TDefR (S n)) ->
```
fredrikNordvallForsberg, line 229: you could do all these for non-`[]` specialisations if you ask for the corresponding data for the specialisation... but for now maybe its enough to note this in a comment

src/Typedefs/Idris.idr
andrevidela: I tried to change it but it messes up with type inference for `Inn`. I'll leave it as is for now, since It will almost certainly change again once I implement `TermWrite` and `TermParse`

src/Typedefs/Idris.idr
andrevidela: Actually I tried and there was a problem because of the substitution mechanism in `TApp` with `ap`. Maybe we don't need to do the substitution? I'll put a comment about that.

andrevidela:
Thanks for your feedback @fredrikNordvallForsberg I'm happy it seems sensible to you too. I started working on `TermWrite` and `TermParse` (on the branch https://github.com/typedefs/typedefs/tree/spec-ctxt-write ) and a huge refactor is coming up, we could certainly push the changes to all other backends after that. It willl make specialisation a lot less hacky and much easier to generalise since every backend needs its only set of specialised types

I'm just waiting for @clayrat 's review and then we can merge

